PR #5, opened 2020-09-23
Add impulse response transformation

francislata:
This pull request adds the `ImpulseResponse` transformation to the list of transformations. This uses the `torch` way of convolving input signal with IR. As a result, batch size and GPU support are added.

Note that in order to support batch size, the definition of a multi-channel has been [edited](https://github.com/asteroid-team/torch-audiomentations/compare/asteroid-team:master...francislata:feature/add-impulse-response-transform?expand=1#diff-cc224b2c34f4db0a2f2e33cc1865595cR29) so that it has the format **(batch size, number of channels, sequence length)**. This is the only modification made to the author's original implementation as to allow `ImpulseResponse` transformation to work properly.

Lastly, the tests checks to ensure that lengths of mixed and original audios do not match as the convolution itself has been tested by the original author.

### DIFF
--- a/torch_audiomentations/augmentations/impulse_response.py
+++ b/torch_audiomentations/augmentations/impulse_response.py
@@ -23,5 +23,9 @@ class ApplyImpulseResponse(BasicTransform):
 
     def forward(self, samples, sample_rate):
         super(ApplyImpulseResponse, self).forward(samples, sample_rate)
-        ir = torch.from_numpy(load_audio(self.parameters["ir_file_path"], sample_rate))
-        return convolve(samples, ir, mode=self.convolve_mode)
+
+        if self.parameters["should_apply"] and len(samples) > 0:
+            ir = torch.from_numpy(load_audio(self.parameters["ir_file_path"], sample_rate))
+            return convolve(samples, ir, mode=self.convolve_mode)
+
+        return samples
--- a/torch_audiomentations/augmentations/impulse_response.py
+++ b/torch_audiomentations/augmentations/impulse_response.py
@@ -18,15 +18,10 @@ class ApplyImpulseResponse(BasicTransform):
 
     def randomize_parameters(self, samples, sample_rate):
         super(ApplyImpulseResponse, self).randomize_parameters(samples, sample_rate)
-
         if self.parameters["should_apply"]:
             self.parameters["ir_file_path"] = random.choice(self.ir_path)
 
-    def apply(self, samples, sample_rate):
-        device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
-        samples = samples.to(device)
-        ir = torch.from_numpy(load_audio(self.parameters["ir_file_path"], sample_rate)).to(device)
-        signal_ir = convolve(samples, ir, mode=self.convolve_mode)
-        signal_ir = signal_ir.cpu().numpy()
-        return signal_ir
-
+    def forward(self, samples, sample_rate):
+        super(ApplyImpulseResponse, self).forward(samples, sample_rate)
+        ir = torch.from_numpy(load_audio(self.parameters["ir_file_path"], sample_rate))
+        return convolve(samples, ir, mode=self.convolve_mode)
--- a/torch_audiomentations/augmentations/impulse_response.py
+++ b/torch_audiomentations/augmentations/impulse_response.py
@@ -1,9 +1,9 @@
 import numpy as np
 import random
 import torch
-from torch_audiomentations.core.transforms_interface import BasicTransform
-from torch_audiomentations.utils.convolution import convolve
-from torch_audiomentations.utils.file import find_files, load_audio
+from ..core.transforms_interface import BasicTransform
+from ..utils.convolution import convolve
+from ..utils.file import find_audio_files, load_audio
 
 
 class ApplyImpulseResponse(BasicTransform):
@@ -13,7 +13,7 @@ class ApplyImpulseResponse(BasicTransform):
 
     def __init__(self, ir_path, convolve_mode="full", p=0.5):
         super(ApplyImpulseResponse, self).__init__(p)
-        self.ir_path = find_files(ir_path)
+        self.ir_path = find_audio_files(ir_path)
         self.convolve_mode = convolve_mode
 
     def randomize_parameters(self, samples, sample_rate):
--- a/torch_audiomentations/augmentations/impulse_response.py
+++ b/torch_audiomentations/augmentations/impulse_response.py
@@ -6,31 +6,26 @@ from torch_audiomentations.utils.convolution import convolve
 from torch_audiomentations.utils.file import find_files, load_audio
 
 
-class ImpulseResponse(BasicTransform):
+class ApplyImpulseResponse(BasicTransform):
     """
     Convolves an input signal with an impulse response.
     """
 
     def __init__(self, ir_path, convolve_mode="full", p=0.5):
-        super(ImpulseResponse, self).__init__(p)
+        super(ApplyImpulseResponse, self).__init__(p)
         self.ir_path = find_files(ir_path)
         self.convolve_mode = convolve_mode
 
     def randomize_parameters(self, samples, sample_rate):
-        super(ImpulseResponse, self).randomize_parameters(samples, sample_rate)
+        super(ApplyImpulseResponse, self).randomize_parameters(samples, sample_rate)
 
         if self.parameters["should_apply"]:
             self.parameters["ir_file_path"] = random.choice(self.ir_path)
 
     def apply(self, samples, sample_rate):
         device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+        samples = samples.to(device)
         ir = torch.from_numpy(load_audio(self.parameters["ir_file_path"], sample_rate)).to(device)
-
-        if isinstance(samples, np.ndarray):
-            samples = torch.from_numpy(samples).to(device)
-        else:
-            samples = samples.to(device)
-
         signal_ir = convolve(samples, ir, mode=self.convolve_mode)
         signal_ir = signal_ir.cpu().numpy()
         return signal_ir
--- a/torch_audiomentations/augmentations/impulse_response.py
+++ b/torch_audiomentations/augmentations/impulse_response.py
@@ -1,0 +1,37 @@
+import numpy as np
+import random
+import torch
+from torch_audiomentations.core.transforms_interface import BasicTransform
+from torch_audiomentations.utils.convolution import convolve
+from torch_audiomentations.utils.file import find_files, load_audio
+
+
+class ImpulseResponse(BasicTransform):
+    """
+    Convolves an input signal with an impulse response.
+    """
+
+    def __init__(self, ir_path, convolve_mode="full", p=0.5):
+        super(ImpulseResponse, self).__init__(p)
+        self.ir_path = find_files(ir_path)
+        self.convolve_mode = convolve_mode
+
+    def randomize_parameters(self, samples, sample_rate):
+        super(ImpulseResponse, self).randomize_parameters(samples, sample_rate)
+
+        if self.parameters["should_apply"]:
+            self.parameters["ir_file_path"] = random.choice(self.ir_path)
+
+    def apply(self, samples, sample_rate):
+        device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+        ir = torch.from_numpy(load_audio(self.parameters["ir_file_path"], sample_rate)).to(device)
+
+        if isinstance(samples, np.ndarray):
+            samples = torch.from_numpy(samples).to(device)
+        else:
+            samples = samples.to(device)
+
+        signal_ir = convolve(samples, ir, mode=self.convolve_mode)
+        signal_ir = signal_ir.cpu().numpy()
+        return signal_ir
+
--- a/torch_audiomentations/utils/file.py
+++ b/torch_audiomentations/utils/file.py
@@ -1,0 +1,21 @@
+import os
+import glob
+import librosa
+
+
+SUPPORTED_EXTENSIONS = ['.wav']
+
+
+def find_files(path):
+    """Finds all files of supported extensions."""
+    files = []
+
+    for supported_extension in SUPPORTED_EXTENSIONS:
+        files.extend(glob.glob(os.path.join(path, '*' + supported_extension)))
+
+    return files
+
+def load_audio(audio_file_path, sample_rate):
+    """Loads the audio given the path of an audio file."""
+    return librosa.load(audio_file_path, sr=sample_rate)[0]
+
--- a/torch_audiomentations/utils/file.py
+++ b/torch_audiomentations/utils/file.py
@@ -18,4 +18,3 @@ def find_audio_files(path):
 def load_audio(audio_file_path, sample_rate):
     """Loads the audio given the path of an audio file."""
     return librosa.load(audio_file_path, sr=sample_rate)[0]
-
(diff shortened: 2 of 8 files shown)


torch_audiomentations/augmentations/impulse_response.py
```diff
@@ -0,0 +1,37 @@
...
+        self.convolve_mode = convolve_mode
+
+    def randomize_parameters(self, samples, sample_rate):
+        super(ImpulseResponse, self).randomize_parameters(samples, sample_rate)
+
+        if self.parameters["should_apply"]:
+            self.parameters["ir_file_path"] = random.choice(self.ir_path)
+
+    def apply(self, samples, sample_rate):
+        device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
```
iver56, line 26: I think `samples` should already be a torch tensor when passed to apply, so we can chain multiple transforms without having to convert between torch and numpy. I also think that it should return a torch tensor, not a numpy ndarray

iver56 (changes requested):

iver56:
I've been thinking about how we should name this transform, since I don't think `AddImpulseResponse` or `ImpulseResponse` is ideal. Some candidates:

* ApplyImpulseResponse
* ConvolveWithRoomImpulseResponse
* ConvolveWithRIR

Any other ideas? @mpariente @hbredin

torch_audiomentations/augmentations/impulse_response.py
francislata: @iver56 - Just addressed the comments for: 1) `samples` being a `torch` tensor, and 2) switching from `ImpulseResponse` to `ApplyImpulseResponse`.

Tests and other files are updated to support these changes.

torch_audiomentations/augmentations/impulse_response.py
```diff
@@ -0,0 +1,32 @@
+import numpy as np
+import random
+import torch
+from torch_audiomentations.core.transforms_interface import BasicTransform
+from torch_audiomentations.utils.convolution import convolve
+from torch_audiomentations.utils.file import find_files, load_audio
```
mpariente, line 6: Shouldn't imports be relative?

torch_audiomentations/utils/file.py
```diff
@@ -0,0 +1,21 @@
...
+    files = []
+
+    for supported_extension in SUPPORTED_EXTENSIONS:
+        files.extend(glob.glob(os.path.join(path, '*' + supported_extension)))
+
+    return files
+
+def load_audio(audio_file_path, sample_rate):
+    """Loads the audio given the path of an audio file."""
+    return librosa.load(audio_file_path, sr=sample_rate)[0]
```
mpariente, line 20: I wouldn't force the sampling rate during loading. 
We can use `soundfile` or `torchaudio` (that we will need eventually) for loading, rather than librosa.

torch_audiomentations/utils/file.py
```diff
@@ -0,0 +1,21 @@
+import os
+import glob
+import librosa
+
+
+SUPPORTED_EXTENSIONS = ['.wav']
+
+
+def find_files(path):
```
mpariente, line 9: The name of the function doesn't seem specific enough. `find_audio_files` maybe?

torch_audiomentations/utils/multichannel.py
```diff
@@ -4,4 +4,4 @@ def is_multichannel(samples):
     :param samples:
     :return:
     """
-    return len(samples.shape) > 1
+    return len(samples.shape) > 2
```
mpariente, line 7: Just to be sure, this assumes that we expect the batch size to be there? 
Just thinking out loud but if the transforms are applied in the `Dataset`, they're applied to individual examples, without batch dim, right?

torch_audiomentations/augmentations/impulse_response.py
```diff
@@ -0,0 +1,32 @@
...
+        self.convolve_mode = convolve_mode
+
+    def randomize_parameters(self, samples, sample_rate):
+        super(ApplyImpulseResponse, self).randomize_parameters(samples, sample_rate)
+
+        if self.parameters["should_apply"]:
+            self.parameters["ir_file_path"] = random.choice(self.ir_path)
+
+    def apply(self, samples, sample_rate):
+        device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
```
mpariente, line 26: I wouldn't force the device to be GPU if there is one. 
We'll probably inherit from `nn.Module` at one point (right?) and can handle device placement after that.

mpariente (commented):
The direction seems good but this PR raises questions that we haven't answered yet IMO, and this is good !

torch_audiomentations/utils/multichannel.py
francislata: It really depends. I would imagine applying augmentations should be done when iterating through a `DataLoader` just before it goes to the model for training, for example. At that time, a `torch.Tensor` input is already present and if the batch size is set to 1, the first dimension will be present.

I think even if it is a mono audio signal, the number of channels dimension should be set to 1 and can be removed by applying `squeeze` to the tensor.

torch_audiomentations/augmentations/impulse_response.py
francislata: I agree. I had questions myself about this so I'm glad you pointed it out. Basically, if we make all transforms inherit from PyTorch's `Module` class, then the user applying transforms can decide to place it to a `torch.device` of their choice.

This change should be pretty easy to do.

torch_audiomentations/utils/file.py
iver56: @mpariente Would you instead want the code to raise a warning if the sample rate of the loaded file doesn't match the expected sample rate?

iver56: Alternatively, raise a warning (about how resampling hurts performance) and then resample? I think I would prefer that

torch_audiomentations/augmentations/impulse_response.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def __init__(self, ir_path, convolve_mode="full", p=0.5):
+        super(ApplyImpulseResponse, self).__init__(p)
+        self.ir_path = find_audio_files(ir_path)
+        self.convolve_mode = convolve_mode
+
+    def randomize_parameters(self, samples, sample_rate):
+        super(ApplyImpulseResponse, self).randomize_parameters(samples, sample_rate)
+        if self.parameters["should_apply"]:
+            self.parameters["ir_file_path"] = random.choice(self.ir_path)
```
iver56, line 22: If we're going to support augmenting batches of sounds, we might want to use different random settings for each sound, right? In that case, I think we're going to need a vector of IR file paths here, and not just a single one

torch_audiomentations/augmentations/impulse_response.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def forward(self, samples, sample_rate):
+        super(ApplyImpulseResponse, self).forward(samples, sample_rate)
+        ir = torch.from_numpy(load_audio(self.parameters["ir_file_path"], sample_rate))
```
iver56, line 26: The transform should only be applied `if self.parameters["should_apply"] and len(samples) > 0`. This relates to the `p` argument. Maybe add a test for this?

torch_audiomentations/utils/file.py
mpariente: Yes, that does make sense. But then, should we load and resample with torchaudio?

iver56: Sure, I'm okay with that. It can also be the scope of a different pull request.

iver56: https://github.com/asteroid-team/torch-audiomentations/issues/9

torch_audiomentations/utils/multichannel.py
mpariente: From what I've seen, the transforms are applied in the `__getitem__` of the `Dataset` (e.g. in albumentations or torchvision). This means that the transform is independent for every element in the batch, and that there is supposedly no batch size when applying the transform.
I think I would agree to have `(batch, 1, samples)` for mono audio because that's what is expected by Conv1D, and it solves one ambiguity. 
And then, we could have `is_multichannel` and `is_batched_multichannel`.

torch_audiomentations/augmentations/impulse_response.py
iver56: https://github.com/asteroid-team/torch-audiomentations/issues/7

mpariente: This makes me realize that you plan to apply augmentations to batches of sounds, instead of individual sounds. Is this really what we want? In this case, do you handle the probabilistic side of augmentations per-batch or per-example?

torch_audiomentations/utils/multichannel.py
iver56: For the sake of limiting the scope of this PR, I'd say we can assume 3 dimensions in the input/output. That way, processing single examples can be supported with batch size = 1. We should definitely keep options open for re-evaluating/changing this afterwards though.

torch_audiomentations/augmentations/impulse_response.py
iver56: Yes, I think that makes sense performance-wise. [The plot](https://github.com/asteroid-team/torch-audiomentations/blob/master/images/convolve_exec_time_plot.png) where I crudely measured batched augmentation vs individual sounds (AKA batch size = 1) showed a 33% reduction in computation time. Further gains are possible with better GPUs.

mpariente: Do you still observe the performance gain if the IRs are different? 
I agree that this _should_ be a possibility, for performance reasons. But it should be possible to apply independent augmentations in the Dataset (which is done in parallel anyway)

iver56: Right. Many people use multiple workers for the dataset generator. How well is GPU supported in such a multi-worker setup? Will it cause problems?

https://pytorch.org/docs/stable/notes/multiprocessing.html#cuda-in-multiprocessing

torch_audiomentations/augmentations/impulse_response.py
iver56: If it would be up to me, I would use absolute imports because I let my IDE (Jetbrains IntelliJ) take care of adding imports (via a keyboard shortcut), and it uses absolute imports.

Also, a quick google search returns this:

bilde

iver56: I'll have a look at some other serious python libraries and see what they do

iver56: Stripe: Absolute imports https://github.com/stripe/stripe-python/blob/master/stripe/api_resources/bank_account.py#L5

Django: Absolute imports 
https://github.com/django/django/blob/master/django/core/paginator.py#L6

Scikit-learn: Relative imports https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/impute/_iterative.py#L11

Keras: Relative imports
https://github.com/keras-team/keras/blob/fb7f49ef5b07f2ceee1d2d6c45f273df6672734c/keras/losses.py#L9

Pandas: Absolute imports https://github.com/pandas-dev/pandas/blob/master/pandas/core/aggregation.py#L23

Flask: Relative imports https://github.com/pallets/flask/blob/master/src/flask/app.py#L32

Looks like a 50/50 to me ^^

mpariente: Those are for the old relative imports, without the `.`. 
The new relative imports use the `.`, they are called explicit relative imports and the part of PEP8 that discourages relative imports has been removed 

Quote from Guido himself
> Isn't this mostly historical? Until the new relative-import syntax was
implemented there were various problems with relative imports. The
short-term solution was to recommend not using them. The long-term
solution was to implement an unambiguous syntax. Now it is time to
withdraw the anti-recommendation. Of course, without going overboard
-- I still find them an acquired taste; but they have their place.

For example, keras, pytorch and albumentations use explicit relative imports. 
pytorch-lightning uses absolute imports. 

I think most of the resources around this debate on Google are quite outdated.

mpariente: Then it's probably fine to keep like this. 
Somehow I mainly wrote code that used explicit relative imports at the beginning and thought this is the norm, happy to know that the other way is fine ! 
Let's stick with this then if you'd like !

iver56: Good point. Well, either way works for me 😄

torch_audiomentations/augmentations/impulse_response.py
iver56: > Do you still observe the performance gain if the IRs are different?

Good question! Indeed the performance gain is smaller when the IRs are different. Crudely measured:

```
torch FFT CUDA       batch size = 1,  single kernel   0.151 s
torch FFT CUDA       batch size = 32, single kernel   0.102 s
torch FFT CUDA       batch size = 32, 32 kernels      0.133 s
```

mpariente: And I don't know how well multiprocessing will behave on GPU.. That would be something to try before making a design choice. Have you looked a albumentations to see how do they do it?

iver56: AFAIK, albumentations is CPU-only. This one seems to support GPU though: https://github.com/kornia/kornia

mpariente: Ok. 
And kornia's augmentations are done out of the dataset in the example. 
https://kornia.readthedocs.io/en/latest/tutorials/data_augmentation.html

francislata: So should we support batches of IRs too? Or we haven't fully decided? Could this be possible as an improvement later on?

iver56: Yes. Sure, we can do it later

iver56: Re kornia's example: Concidentally, that was the way I thought of doing it in my training pipeline. I.e. let the data loader take care of loading data (using CPU only), then do a data augmentation step (on the GPU), then pass to the model. Thanks for verifying that this is a valid approach 😄

iver56 (approved):

iver56:
Thanks!

torch_audiomentations/augmentations/impulse_response.py
```diff
@@ -0,0 +1,31 @@
...
+        self.ir_path = find_audio_files(ir_path)
+        self.convolve_mode = convolve_mode
+
+    def randomize_parameters(self, samples, sample_rate):
+        super(ApplyImpulseResponse, self).randomize_parameters(samples, sample_rate)
+        if self.parameters["should_apply"]:
+            self.parameters["ir_file_path"] = random.choice(self.ir_path)
+
+    def forward(self, samples, sample_rate):
+        super(ApplyImpulseResponse, self).forward(samples, sample_rate)
```
mpariente, line 25: Is this really what we want though? 
This should still be `apply`, which will be called by the `BasicTransform`'s `forward` not? So we avoid calling this line every-time?

iver56: Agreed

francislata: I guess we can do that too. Though I think to stick with the torch way of subclassing `nn.Module` is to override `forward`.

iver56: We'll still be compatible with the `nn.Module`, since the superclass implements forward and calls the subclass' apply :)

francislata: I just noticed that sanity checks for ensuring the input signal is non-empty and if it should apply the augmentation will be eliminated for _all_ subclasses of `BasicTransform`.

PR #12, opened 2020-09-24
[ci] Add GitHub actions to lint/test

mpariente:
- Linting with `pylint` and `black`. We can talk about what exceptions we tolerate. Do we want to lint `scripts` as well? 
- For the unit test config, it's not fully tested but it should work. 
- I'm stuck with the requirements, how should we proceed? `environement.yml` is nice, I like it. But having a `requirements.txt` can be useful as well for non-conda users. What do you think about having `requirements.txt` and `dev-requirements.txt` and having them called in the `environement.yml` (with `-r`)? That would make a nice compromise IMO. 
- Codecov should be set up for asteroid but also not sure yet.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,13 @@
+[flake8]
+max-line-length = 119
+exclude = scripts,*.egg,build
+select = E,W,F
+verbose = 2
+# https://pep8.readthedocs.io/en/latest/intro.html#error-codes
+format = pylint
+ignore =
+    E731  # E731 - Do not assign a lambda expression, use a def
+    W605  # W605 - invalid escape sequence '\_'. Needed for docs
+    W504  # W504 - line break after binary operator
+    W503  # W503 - line break before binary operator, need for black
+    E203  # E203 - whitespace before ':'. Opposite convention enforced by black
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,9 @@
+[report]
+fail_under=80
+# Regexes for lines to exclude from consideration
+exclude_lines =
+  pragma: no cover
+  raise NotImplementedError
+
+[run]
+source=torch_audiomentations
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Install python dependencies
       run: |
         python -m pip install --upgrade --user pip --quiet
-        python -m pip install -r requirements.txt
+        python -m pip install -r ./.github/workflows/test-requirements.txt
         python --version
         pip --version
         python -m pip list
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,11 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: Install libnsdfile
+      run: |
+        sudo apt update
+        sudo apt install libsndfile1-dev libsndfile1
+
     # FIX requirement install
     - name: Install python dependencies
       run: |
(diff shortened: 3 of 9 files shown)


mpariente:
Do you know how conflicts between conda and pip are handled in the env file? e.g if pytorch is both in `dependencies` and `pip`? I guess that pip will just saw requirement satisfied but I'm not sure. In this case, it's fine to repeat numpy/pytorch/scipy in both deps and pip. WDYT?

.flake8
```diff
@@ -0,0 +1,13 @@
+[flake8]
+max-line-length = 119
```
iver56, line 2: Maybe we should align this value with `line-length` in pyproject.toml? Or is this value higher on purpose to give some leeway?

iver56:
> 
> 
> Do you know how conflicts between conda and pip are handled in the env file? e.g if pytorch is both in `dependencies` and `pip`? I guess that pip will just saw requirement satisfied but I'm not sure. In this case, it's fine to repeat numpy/pytorch/scipy in both deps and pip. WDYT?

I've seen errors where the culprit was that numpy was somehow installed twice (probably by both conda and pip). Haven't had any other negative experiences with combining conda and pip though. Is it troublesome to use conda in Github Actions?

.flake8
mpariente: We can. We didn't agree on the line length yet, so I didn't pay attention to it.

mpariente:
My guess is that setting up conda will take quite some time. 
Also, how does it behave to install cudatoolkit on a CPU-only machine?

mpariente:
Maybe I'm missing something. How would you handle the deps specification between 
- conda env file
- requirements.txt
- setup.py
In particular, the versions are currently fixed in the conda env file (`==`), is this necessary? Would you also fix it in the setup.py?

iver56:
> 
> 
> Maybe I'm missing something. How would you handle the deps specification between
> 
>     * conda env file
> 
>     * requirements.txt
> 
>     * setup.py
>       In particular, the versions are currently fixed in the conda env file (`==`), is this necessary? Would you also fix it in the setup.py?

* setup.py should not be very strict when it comes to ranges of versions.
* Currently, environment.yml shows one of many possible combinations of package versions that should work. I think it makes sense to pin packages at least some packages for development and testing.
* I've never worked on a project that had both environment.yml and requirements.txt. The reason why I prefer conda is that it's the only way I've been able to install pytorch with CUDA support on Windows.
* In other packages, I've been handling dependencies in requirements.txt and setup.py separately. E.g. `audiomentations` has additional dependencies in requirements.txt and more strict version specs in requirements.txt compared to setup.py

iver56:
> 
> 
> My guess is that setting up conda will take quite some time.
> Also, how does it behave to install cudatoolkit on a CPU-only machine?

I see. How about instead making a Dockerfile for running tests in Github Actions? That seems more feasible, doen't it?

Re cudatoolkit on a CPU-only machine: I've been doing `RUN sed -i.bak "s|- cudatoolkit=10.1|- cpuonly|" environment.yml` to deal with this

mpariente:
IDK, will we need ffmpeg during testing?

iver56:
> 
> 
> IDK, will we need ffmpeg during testing?

It depends. Do we need to load special audio file formats, like ogg/opus in testing?

mpariente:
No, I don't think we'll ned any of this. That's why I don't think it's worth it to setup conda for CI. 
I'll add a req file in ./github/workflows with all what's needed for testing and we'll see how we handle deps later.

mpariente:
Ready to merge IMO

Issue #11, opened 2020-09-24
BasicTransform subclasses should implement apply, not forward


iver56:
Ref https://github.com/asteroid-team/torch-audiomentations/pull/5#discussion_r494317490

francislata:
I am thinking to address this pretty soon with the new background noise augmentation that I am currently working on.

PR #13, opened 2020-09-25
Add background noise transformation

francislata:
This pull request adds a new transformation for adding background noise. Note that it is required for all input samples to be a `Tensor` type with `size()` of (batch size, sequence_length). In the future, more multi-channel refactoring needs to be done.

There's an **underlying assumption** that goes with this implementation: all input samples must be stacked and have the same sequence length. This assumption is self-explanatory and handles the common use cases of handling augmentations during training:

```python
for x, y in dataloader:
    x, y = x.to(device), y.to(device)
    x = ApplyBackgroundNoise(x)
```

In addition, https://github.com/asteroid-team/torch-audiomentations/issues/11 has been addressed and the `ApplyImpulseResponse` transformation has been updated accordingly.

Lastly, batched IR and background noise are now supported in this pull request to provide more efficiency. Note that if the IR and background noise paths have less files than the input audio, that the random selector picks _with replacement_ from the existing files given the provided path. This ensures that there will be an associated IR and background noise for every input audio.


torch_audiomentations/augmentations/background_noise.py
```diff
@@ -0,0 +1,53 @@
...
+from ..utils.file import find_audio_files, load_audio
+from ..utils.dsp import calculate_rms, calculate_desired_noise_rms
+
+
+class ApplyBackgroundNoise(BasicTransform):
+    """
+    Applies a background noise to the input signal.
+    """
+
+    def __init__(self, bg_path, snr_in_db, p=0.5):
```
mpariente, line 13: Could we support a distribution on `snr_in_db` instead of a fixed number? That would be more useful IMO.

torch_audiomentations/augmentations/background_noise.py
```diff
@@ -0,0 +1,53 @@
...
+        super(ApplyBackgroundNoise, self).__init__(p)
+        self.bg_path = find_audio_files(bg_path)
+        assert len(self.bg_path) > 0
```
mpariente, line 16: Raise a useful error like `Background noise directory is empty`.

torch_audiomentations/augmentations/background_noise.py
```diff
@@ -0,0 +1,53 @@
...
+        if self.parameters["should_apply"] and samples.size(0) > 0:
+            bg_file_paths = random.choices(self.bg_path, k=samples.size(0))
+            bg_audios = []
+            for index, bg_file_path in enumerate(bg_file_paths):
+                bg_audio = torch.from_numpy(load_audio(bg_file_path, sample_rate))
+                max_bg_offset = max(0, bg_audio.size(0) - samples[index].size(0) - 1)
+
+                bg_start_index = random.randint(0, max_bg_offset)
+                bg_end_index = bg_start_index + samples[index].size(0)
+                bg_audios.append(bg_audio[bg_start_index:bg_end_index])
```
mpariente, line 30: In the case where noise files are much longer that the input, this doesn't look very efficient. 
`soundfile` supports accessing the length of the audio without loading and loading only a portion of it, which would be more efficient.

torch_audiomentations/augmentations/background_noise.py
```diff
@@ -0,0 +1,53 @@
...
+        desired_bg_audios_rms = calculate_desired_noise_rms(
+            samples_rms, self.parameters["snr_in_db"]
+        )
+        bg_audios = bg_audios * (desired_bg_audios_rms / bg_audios_rms)
+
+        while bg_audios.size(1) < samples.size(1):
+            bg_audios = bg_audios.repeat(1, 2)
+
+        if bg_audios.size(1) > samples.size(1):
+            bg_audios = bg_audios[:, : samples.size(1)]
```
mpariente, line 51: It seems weird to me to duplicate them in the `apply`, and after computing the scaling. 
Rather draw more in line 22 at the beginning of `randomize_parameters`, so your noise samples don't have to be aligned, and the noise rms corresponds to the desired SNR.

mpariente (commented):
I see now that it's a draft, were you expecting a review anyway? Sorry if not.

torch_audiomentations/augmentations/background_noise.py
francislata: Got it. I think there's a slight misunderstanding with line [22](https://github.com/asteroid-team/torch-audiomentations/pull/13/files/7ce0aab98821538c9c74298ac52daec9d133536a#diff-e9d818488a0e110d9441b5ec68f34eafR22) because that line draws multiple samples to ensure that the number of background noises matches the number of samples to process.

mpariente: So why do you need to repeat and truncate? I don't understand.

francislata:
The idea of repeating and truncating was to ensure that if the background noise is shorter than the input samples, then we can handle it by doing that process of repeating and truncating

--
Francis Lata

torch_audiomentations/augmentations/background_noise.py
francislata: Mean to respond to your answer but got added as a comment instead, @mpariente - here's my response: https://github.com/asteroid-team/torch-audiomentations/pull/13#issuecomment-699935666

torch_audiomentations/augmentations/background_noise.py
francislata: Resolving at this moment since latest commits take advantage of `soundfile`'s audio chunk loading to optimize the performance. Conditions are checked whether to repeat and truncate or not so that the appropriate steps are performed.

iver56:
test_fixtures/bg/Single_403a_2_Noise_Babble.wav seems too large to include in the repository. In my opinion, test fixtures / demo files should be <500 kb. It can be a little bit tricky to replace the file, since you have to kind of rewrite the commit that added the file, or else it'll stay stuck in the commit history. Perhaps the easiest way to do it would be to squash the commits and then amend the combined commit with a smaller file, and then force-push.

torch_audiomentations/augmentations/background_noise.py
mpariente: Ok, I misunderstood the code sorry. 
Thanks.

torch_audiomentations/utils/dsp.py
```diff
@@ -0,0 +1,32 @@
...
+    :return:
+    """
+    a = float(snr) / 20
+    noise_rms = clean_rms / (10 ** a)
+    return noise_rms
+
+
+def resample_audio(audio, orig_sr, target_sr):
+    """Resamples the audio to a new sampling rate."""
+    return librosa.resample(audio, orig_sr, target_sr)
```
mpariente, line 32: Why is this needed? 
We can call `librosa.resample` directly right?
This way, when we'll read the code, we'll know which resampling we use.

torch_audiomentations/utils/file.py
```diff
@@ -16,6 +17,11 @@ def find_audio_files(path):
     return files
 
 
-def load_audio(audio_file_path, sample_rate):
+def load_audio(audio_file_path, sample_rate=None, start=0, stop=None):
```
mpariente, line 20: :+1:

torch_audiomentations/utils/dsp.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    return torch.sqrt(torch.mean(torch.square(samples), -1, keepdims=True))
+
+
+def calculate_desired_noise_rms(clean_rms, snr):
+    """
+    Given the Root Mean Square (RMS) of a clean sound and a desired signal-to-noise ratio (SNR),
+    calculate the desired RMS of a noise sound to be mixed in.
+    Based on https://github.com/Sato-Kunihiko/audio-SNR/blob/8d2c933b6c0afe6f1203251f4877e7a1068a6130/create_mixed_audio_file.py#L20
+                Taken from https://github.com/iver56/audiomentations/blob/master/audiomentations/core/utils.py
```
mpariente, line 20: Which of those two?

torch_audiomentations/augmentations/impulse_response.py
```diff
@@ -11,21 +11,36 @@ class ApplyImpulseResponse(BasicTransform):
...
-            return convolve(samples, ir, mode=self.convolve_mode)
+        if self.parameters["should_apply"] and samples.size(0) > 0:
+            ir_paths = random.choices(self.ir_path, k=samples.size(0))
+            ir_sounds = []
+            max_ir_sound_length = 0
+            for ir_path in ir_paths:
+                ir_samples = load_audio(ir_path, sample_rate)
+                num_samples = len(ir_samples)
+                if num_samples > max_ir_sound_length:
+                    max_ir_sound_length = num_samples
```
mpariente, line 34: A bit clearer I guess. 
```suggestion
                max_ir_sound_length = max(len(ir_samples), max_ir_sound_length)
```

torch_audiomentations/utils/file.py
```diff
@@ -16,6 +17,11 @@ def find_audio_files(path):
...
     """Loads the audio given the path of an audio file."""
-    return librosa.load(audio_file_path, sr=sample_rate)[0]
+    audio, source_sample_rate = soundfile.read(audio_file_path, start=start, stop=stop)
+
+    if sample_rate:
+        audio = resample_audio(audio, source_sample_rate, sample_rate)
+
+    return audio
```
mpariente, line 27: The sample rate should probably be returned. I guess we'll want it/ need it when it's `None` right?

torch_audiomentations/__init__.py
```diff
@@ -1,6 +1,8 @@
 from .augmentations.polarity_inversion import PolarityInversion
 from .augmentations.impulse_response import ApplyImpulseResponse
+from .augmentations.background_noise import ApplyBackgroundNoise
 from .utils.convolution import convolve
 from .utils.file import find_audio_files, load_audio
+from .utils.dsp import calculate_rms, calculate_desired_noise_rms, resample_audio
```
mpariente, line 6: I would limit the functions imported in `__init__` to what will be used by the user. @iver56 ?

torch_audiomentations/utils/file.py
```diff
@@ -16,6 +17,11 @@ def find_audio_files(path):
     return files
 
 
-def load_audio(audio_file_path, sample_rate):
+def load_audio(audio_file_path, sample_rate=None, start=0, stop=None):
     """Loads the audio given the path of an audio file."""
-    return librosa.load(audio_file_path, sr=sample_rate)[0]
+    audio, source_sample_rate = soundfile.read(audio_file_path, start=start, stop=stop)
```
mpariente, line 22: Another question I don't have the answer to. 
Should start and stop be wrt the specified sample rate or the source sample rate? 
It seems more natural to consider the specified sample rate IMO.

torch_audiomentations/augmentations/background_noise.py
```diff
@@ -0,0 +1,92 @@
...
+                    )
+                    bg_audio = [current_bg_audio]
+
+                    while loaded_bg_audio_num_samples < samples_num_samples:
+                        current_bg_audio = bg_audio[-1]
+                        loaded_bg_audio_num_samples += current_bg_audio.shape[0]
+                        bg_audio.append(current_bg_audio)
+
+                    bg_audio = np.concatenate(bg_audio)
+                    bg_audio = bg_audio[:samples_num_samples]
```
mpariente, line 61: Make this a function `repeat_until_len` or something like that. The whole `randomize_parameters` will be more readable.

torch_audiomentations/augmentations/background_noise.py
```diff
@@ -0,0 +1,92 @@
...
+            bg_audios = []
+            for index, bg_file_path in enumerate(bg_file_paths):
+                bg_audio_info = soundfile.info(bg_file_path, verbose=True)
+                bg_audio_num_samples = math.ceil(
+                    sample_rate * bg_audio_info.frames / bg_audio_info.samplerate
+                )
+                samples_num_samples = samples[index].size(0)
+
+                # ensure that background noise has the same length as the sample
+                if bg_audio_num_samples < samples_num_samples:
```
mpariente, line 42:
```suggestion
                if bg_audio_num_samples <= samples_num_samples:
```

torch_audiomentations/augmentations/background_noise.py
```diff
@@ -0,0 +1,92 @@
...
+                    bg_audio = load_audio(
+                        bg_file_path,
+                        sample_rate=sample_rate,
+                        start=bg_start_index,
+                        stop=bg_stop_index,
+                    )
+
+                bg_audios.append(torch.from_numpy(bg_audio))
+
+            self.parameters["snr_in_db"] = random.uniform(self.min_snr_in_db, self.max_snr_in_db)
```
mpariente, line 76: That'll make one SNR for the whole batch. This is not what we want I guess. We want variability in the batch.

mpariente (commented):
I still have some remarks and questions. Don't take everything I say as "true" but rather let's discuss some of the choices. 

I think having a distribution for SNR would be quite nice. And each person will want to have a different one, maybe gaussian, maybe uniform etc.. 
I'm wondering if we shouldn't use `eval` for it so everyone could choose what they want. 
It would look like this: 
```python
# Easy normal way
ApplyBackgroundNoise(path, snr=-3)  
# Choose your distribution
ApplyBackgroundNoise(path, snr="np.random.randn()")
# Maybe this as well for uniform? Don't know
ApplyBackgroundNoise(path, snr=[-6, 3])  
```

WDYT?

torch_audiomentations/utils/file.py
iver56: Good point. As it stands, it's also not entirely clear if start and stop is in samples or in seconds

torch_audiomentations/utils/dsp.py
iver56: Actually, audiomentations took it from audio-SNR. We can keep the audio-SNR link and remove the audiomentations one.

torch_audiomentations/__init__.py
iver56: Yes, that makes sense. I've recently done the same thing in audiomentations. If some users want to use our internal utils, they can do a workaround, like copying the code or doing a deeper import (and taking the risk that internals might change across versions).

iver56:
> 
> 
> I still have some remarks and questions. Don't take everything I say as "true" but rather let's discuss some of the choices.
> 
> I think having a distribution for SNR would be quite nice. And each person will want to have a different one, maybe gaussian, maybe uniform etc..
> I'm wondering if we shouldn't use `eval` for it so everyone could choose what they want.
> It would look like this:
> 
> ```python
> # Easy normal way
> ApplyBackgroundNoise(path, snr=-3)  
> # Choose your distribution
> ApplyBackgroundNoise(path, snr="np.random.randn()")
> # Maybe this as well for uniform? Don't know
> ApplyBackgroundNoise(path, snr=[-6, 3])  
> ```
> 
> WDYT?

I think we can pick a random snr (in dB) for now, and create an issue where we discuss future improvements, to limit the scope of this pull request.

Also, this Twitter thread is relevant
https://twitter.com/keunwoochoi/status/1301954734690127872

mpariente:
Ok for limiting this PR.

torch_audiomentations/utils/dsp.py
```diff
@@ -0,0 +1,32 @@
...
+import torch
+
+
+def calculate_rms(samples):
+    """
+    Calculates the root mean square.
+
+                Taken from https://github.com/iver56/audiomentations/blob/master/audiomentations/core/utils.py
+    """
+    return torch.sqrt(torch.mean(torch.square(samples), -1, keepdims=True))
```
iver56, line 12:
```suggestion
    return torch.sqrt(torch.mean(torch.square(samples), -1, keepdim=True))
```

ref https://pytorch.org/docs/stable/generated/torch.mean.html

iver56:
I'd like to carry through this pull request soon, since it contains some contributions to the core that further development depends on

francislata:
@iver56 - I apologize for being inactive these past few days due to other responsibilities. I'm expecting a turn around in the next couple of days. I haven't gone through the comments since last week. What are the core changes required to get this PR approved?

iver56:
> 
> 
> @iver56 - I apologize for being inactive these past few days due to other responsibilities. I'm expecting a turn around in the next couple of days. I haven't gone through the comments since last week. What are the core changes required to get this PR approved?

That's okay :) I guess the most important changes are:

* Replace test_fixtures/bg/Single_403a_2_Noise_Babble.wav with a shorter variant (and make sure the original huge file doesn't get included in the git history)
* Fix the keepdims typo

There are also some other remarks (see comments above ^), but we could maybe save some of them for later

iver56:
Thanks for the contribution! I might polish some edges later

PR #15, opened 2020-10-08
Enable variability within a batch

iver56:
Torch-audiomentations should support batch processing. Each batch contains _n_ (potentially multichannel) audio snippets.

Each audio snippet within a batch should processed with potentially different parameters. Furthermore, some of the audio snippets within a batch should remain unprocessed (this can be controlled with the `p` value). The base transform class now does a pre-selection of which audio snippets should be processed, and the implemented apply_transform would only work on those.

TODO:
* [x] Adapt and test AddBackgroundNoise
* [x] Adapt and test AddImpulseResponse

Question: How should we handle transforms that change the length of each audio snippet differently? Calculate the max audio snippet length and zero-pad all other audio snippets accordingly?

Note: This pull request also adds some TODOs that remain from #13 

Note: The reason why I renamed apply to apply_transform is that apply is already a thing in nn.Module. Didn't want to interfere with that.

### DIFF
--- a/torch_audiomentations/augmentations/background_noise.py
+++ b/torch_audiomentations/augmentations/background_noise.py
@@ -32,10 +32,14 @@ class ApplyBackgroundNoise(BaseWaveformTransform):
 
         self.min_snr_in_db = min_snr_in_db
         self.max_snr_in_db = max_snr_in_db
+        self.snr_distribution = torch.distributions.Uniform(
+            low=min_snr_in_db, high=max_snr_in_db
+        )
         self.device = device
 
     def randomize_parameters(self, selected_samples, sample_rate: int):
-        bg_file_paths = random.choices(self.bg_path, k=selected_samples.size(0))
+        batch_size = selected_samples.size(0)
+        bg_file_paths = random.choices(self.bg_path, k=batch_size)
         bg_audios = []
         for index, bg_file_path in enumerate(bg_file_paths):
             bg_audio_info = soundfile.info(bg_file_path, verbose=True)
@@ -83,8 +87,8 @@ class ApplyBackgroundNoise(BaseWaveformTransform):
 
             bg_audios.append(torch.from_numpy(bg_audio))
 
-        self.parameters["snr_in_db"] = random.uniform(
-            self.min_snr_in_db, self.max_snr_in_db
+        self.parameters["snr_in_db"] = self.snr_distribution.sample(
+            sample_shape=(batch_size,)
         )
         self.parameters["bg_audios"] = torch.stack(bg_audios)
 
@@ -93,12 +97,12 @@ class ApplyBackgroundNoise(BaseWaveformTransform):
         bg_audios = self.parameters["bg_audios"].to(self.device)
 
         # calculate sample and background audio RMS
-        samples_rms = calculate_rms(selected_samples)
-        bg_audios_rms = calculate_rms(bg_audios)
+        samples_rms = calculate_rms(selected_samples).squeeze()
+        bg_audios_rms = calculate_rms(bg_audios).squeeze()
 
         desired_bg_audios_rms = calculate_desired_noise_rms(
             samples_rms, self.parameters["snr_in_db"]
         )
-        bg_audios = bg_audios * (desired_bg_audios_rms / bg_audios_rms)
+        bg_audios = bg_audios * (desired_bg_audios_rms / bg_audios_rms).unsqueeze(1)
 
         return selected_samples + bg_audios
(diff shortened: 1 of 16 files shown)


mpariente:
> Question: How should we handle transforms that change the length of each audio snippet differently? Calculate the max audio snippet length and zero-pad all other audio snippets accordingly?

What else than speed perturbation would change the length of wav tensors? Padding seems ok I guess. I'd rather use very low energy noise than synthetic zeros for stability.

torch_audiomentations/core/transforms_interface.py
```diff
@@ -13,45 +14,87 @@ class EmptyPathException(Exception):
...
+                    sample_shape=(batch_size,)
+                ).to(torch.bool)
+            }
+            if self.parameters["should_apply"].any():
+                self.randomize_parameters(
+                    samples[self.parameters["should_apply"]], sample_rate
+                )
+
+        if self.parameters["should_apply"].any():
+            cloned_samples = samples.detach().clone()
```
mpariente, line 74: Are we sure we want to detach here?

torch_audiomentations/augmentations/impulse_response.py
```diff
@@ -21,26 +26,26 @@ def __init__(self, ir_path, device=torch.device("cpu"), convolve_mode="full", p=
...
-        samples = samples.to(self.device)
+    def randomize_parameters(self, selected_samples, sample_rate: int):
+        ir_paths = random.choices(self.ir_path, k=selected_samples.size(0))
+        ir_sounds = []
+        max_ir_sound_length = 0
+        for ir_path in ir_paths:
+            ir_samples = load_audio(ir_path, sample_rate)
+            num_samples = len(ir_samples)
+            if num_samples > max_ir_sound_length:
+                max_ir_sound_length = num_samples
```
mpariente, line 37:
```suggestion
            max_ir_sound_length = max(max_ir_sound_length, len(ir_samples))
```

torch_audiomentations/augmentations/background_noise.py
```diff
@@ -26,67 +34,71 @@ def __init__(
...
+                    sample_rate=sample_rate,
+                    start=bg_start_index,
+                    stop=bg_stop_index,
+                )
+
+            bg_audios.append(torch.from_numpy(bg_audio))
+
+        self.parameters["snr_in_db"] = random.uniform(
+            self.min_snr_in_db, self.max_snr_in_db
+        )
```
mpariente, line 88: Should probably be a tensor. 
But again, that can come in a later PR.

mpariente (commented):
Overall, this looks good. (the diff looks weird though) 
Did you want feedback on a particular point?

torch_audiomentations/core/transforms_interface.py
iver56: Not sure, but this seemed to be the recommended way of cloning a tensor in pytorch
https://stackoverflow.com/questions/55266154/pytorch-preferred-way-to-copy-a-tensor

I thought I needed to clone it to avoid changing the original tensor, ref discussion on Slack

mpariente: But this removes the graph. 
`kornia` advertises for differentiability, I don't know what's their main use case, but if we do it in PyTorch, we might want to keep things backpropagable as well. 
The first use case I think about is a TasNet-like architecture with some kind of SpecAugment on the learned features, you won't want to detach in this case.

iver56: Yeah okay, that's interesting. I guess I could remove the detach and keep the clone, then. Btw, how do we test (as in unit test) differentiability and check that the graph is kept intact?

mpariente: Yes.

For differentiability, you can backward anything that depends on the input and check that the gradients are not `None` anymore (with `requires_grad` inputs). Regarding the graph, I wouldn't unit test that, it's pytorch's job to handle it. You cannot change the graph on nodes "before you" anyway. 

We can use `with torch.autograd.detect_anomaly():` to make sure there is no NaN either.

iver56:
> 
> 
> > Question: How should we handle transforms that change the length of each audio snippet differently? Calculate the max audio snippet length and zero-pad all other audio snippets accordingly?
> 
> What else than speed perturbation would change the length of wav tensors? Padding seems ok I guess. I'd rather use very low energy noise than synthetic zeros for stability.

Different impulse response lengths lead to different output lengths. However, I simply cropped them to the original length in this pull request. It seemed appropriate, as it's what I do in a different code base already. If anyone needs to keep the whole reverb tail at the end, they can open a pull request that adds a parameter, or they can call convolve manually ;)

PR #24, opened 2020-10-19
feat: turn transform into identity when `training` is False

### LINKED ISSUE #22: Add `train()` method?
Suppose I define `MyNetwork` with built-in augmentation like this:

```python
import torch
import torch.nn as nn
from torch_audiomentations import Gain

class MyNetwork(nn.Module):
    def __init__(self):
        self.augmentation = Gain(min_gain_in_db=-15.0, max_gain_in_db=5.0, p=0.5)
        self.other_layers = ...

    def foward(self, waveforms):
        augmented_waveforms = self.augmentation(waveforms)
        return self.other_layers(augmented_waveforms)

net = MyNetwork()
```        

As per my understanding of current code base, calling `net.train(mode=False)` (aka `net.eval()`) has no effect on `net.augmentation`. Augmentation will always be applied, whether we call `net.eval()` or not.

Would it make sense to add the following method to `BaseWaveformTransform` so that one can activate/deactivate augmentations with `net.train(mode=...)`?

```python
class BaseWaveformTransform(nn.Module):
    def train(self, mode: bool = True):
        if mode:
            self.p = self._p_backup
        else:
            self._p_backup = self.p
            self.p = 0.0
```

The `_p_backup` hack is ugly and the whole `p` management should probably be rewritten but you get the idea. What do you think?

hbredin:
Closes #22

### DIFF
--- a/tests/test_gain.py
+++ b/tests/test_gain.py
@@ -30,6 +30,19 @@ class TestGain(unittest.TestCase):
         )
         self.assertEqual(processed_samples.dtype, np.float32)
 
+    def test_eval(self):
+        samples = np.array([[[1.0, 0.5, -0.25, -0.125, 0.0]]], dtype=np.float32)
+        sample_rate = 16000
+
+        augment = Gain(min_gain_in_db=-15, max_gain_in_db=5, p=1.0)
+        augment.eval()
+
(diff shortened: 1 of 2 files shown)


iver56:
Thanks!

PR #31, opened 2020-11-02
feat: add from_dict/from_yaml utility functions

### LINKED ISSUE #25: Configuration file
Would be nice for reproducibility reasons to be able to instantiate augmentations from a configuration file.

```python
augment = AddGaussianNoise(min_amplitude=0.001, max_amplitude=0.015, p=0.5)
```

would be equivalent to loading from this configuration file:

```yaml
# config.yaml
AddGaussianNoise:
    min_amplitude: 0.001
    max_amplitude: 0.015
    p: 0.5
```

```python
with open('config.yaml', 'r') as fp:
    augment = torch_audiomentations.from_config(yaml.load(fp))
```

Composition could have the following (simple) syntax:

```yaml
# config.yaml
Compose:
    AddGaussianNoise:
        min_amplitude: 0.001
        max_amplitude: 0.015
        p: 0.5
    TimeStretch:
        min_rate: 0.8
        max_rate: 1.25
        p: 0.5
```

Would you consider a PR adding such a feature?

hbredin:
Closes #25

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -12,24 +12,16 @@ from torch_audiomentations import Gain
 
 class TestFromConfig(unittest.TestCase):
     def test_from_dict(self):
-        config = {"Gain": {"min_gain_in_db": -12.0, "mode": "per_channel"}}
+        config = {
+            "transform": "Gain",
+            "params": {"min_gain_in_db": -12.0, "mode": "per_channel"},
+        }
         transform = from_dict(config)
 
         assert isinstance(transform, Gain)
         assert transform.min_gain_in_db == -12.0
         assert transform.max_gain_in_db == 6.0
         assert transform.mode == "per_channel"
-
-    # TODO: uncomment once Compose is available
-    # https://github.com/asteroid-team/torch-audiomentations/issues/23
-    # def test_from_dict_compose(self):
-    #     config = {
-    #         "Gain": {"min_gain_in_db": -12.0, "mode": "per_channel"},
-    #         "PolarityInversion": {},
-    #     }
-
-    #     transform = from_dict(config)
-    #     assert isinstance(transform, Compose)
 
     def test_from_yaml(self):
         file_yml = TEST_FIXTURES_DIR / "config.yml"
@@ -39,3 +31,27 @@ class TestFromConfig(unittest.TestCase):
         assert transform.min_gain_in_db == -12.0
         assert transform.max_gain_in_db == 6.0
         assert transform.mode == "per_channel"
+
+    # TODO: uncomment once Compose is available
+    # https://github.com/asteroid-team/torch-audiomentations/issues/23
+
+    # def test_from_dict_compose(self):
+    #     config = {
+    #         "Compose": {
+    #             "shuffle": True,
+    #             "transforms": [
+    #                 {
+    #                     "transform": "Gain",
+    #                     "params": {"min_gain_in_db": -12.0, "mode": "per_channel"},
+    #                 },
+    #                 {"transform": "PolarityInversion"},
+    #             ],
+    #         }
+    #     }
+    #     transform = from_dict(config)
+    #     assert isinstance(transform, Compose)
+
+    # def test_from_yaml_compose(self):
+    #     file_yml = TEST_FIXTURES_DIR / "config_compose.yml"
+    #     transform = from_yaml(file_yml)
+    #     assert isinstance(transform, Compose)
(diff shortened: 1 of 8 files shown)


torch_audiomentations/utils/config.py
```diff
@@ -0,0 +1,126 @@
...
+
+    >>> from torch_audiomentations import TransformClassName
+    >>> transform = TransformClassName(param_name=param_value, ...)
+
+    >>> transform = from_dict({"TransformClassName": {"param_name": param_value, ...}}) 
+
+    Transforms composition is also supported:
+
+    >>> compose = from_dict({"FirstTransform": {"param": value},
+    ...                      "SecondTransform": {"param": value}})
```
iver56, line 40: Hmm, a dict where the order of keys matters. Does that work well in Python 3.6? Perhaps it's better to use an OrderedDict to be on the safe side?

https://stackoverflow.com/a/39980744/2319697

iver56:
Yeah, sorry, `test_requires` isn't actually used in the CI pipeline right now. See .github/workflows/test_requirements.txt. We should probably fix this inconsistency later.

tests/test_config.py
```diff
@@ -0,0 +1,41 @@
...
+        assert isinstance(transform, Gain)
+        assert transform.min_gain_in_db == -12.0
+        assert transform.max_gain_in_db == 6.0
+        assert transform.mode == "per_channel"
+
+    # TODO: uncomment once Compose is available
+    # https://github.com/asteroid-team/torch-audiomentations/issues/23
+    # def test_from_dict_compose(self):
+    #     config = {
+    #         "Gain": {"min_gain_in_db": -12.0, "mode": "per_channel"},
```
iver56, line 27: Perhaps it's better to be more explicit about how we initialize a Compose? Because Compose is going to have arguments like shuffle. And we are going to have OneOf and SomeOf, which are other ways of applying (several) transforms sequentially

torch_audiomentations/utils/config.py
mpariente: Agreed.

tests/test_config.py
mpariente: IMO, it's probably enough to do that for now and adapt when `Compose` is implemented.

torch_audiomentations/utils/config.py
hbredin: How would that work with `from_yaml` for which `yaml.load` gives me a `dict` and not an `OrderedDict`? 
https://github.com/yaml/pyyaml/issues/110

There are two options I can think of:

- switch to a different syntax using a list (but this will make writing YAML more cumbersome)
- simpler but more involved: switch to Python 3.7+ for which Python insertion order is guaranteed (Python is now 3.9 already...)

tests/test_config.py
hbredin: @mpariente What do you mean by **"that"** exactly?

mpariente: Sorry 

"It's probably enough to keep it as is for now and adapt when Compose is implemented."

hbredin: So, no change needed from my side for now?

hbredin: We could also keep this PR open until `Compose` is implemented.

torch_audiomentations/utils/config.py
iver56: What would the syntax with list look like? I think I would prefer that if it means we can stay compatible with 3.6 (which lots of people are still using)

hbredin: Something like:
```yaml
- transform: FirstTransform
  params:
    param1: value1
    param2: value2
- transform: SecondTransform
  params:
    param3: value3
```

that converts to this Python list:

```python
[{'transform': 'FirstTransform',
  'params': {'param1': 'value1', 'param2': 'value2'}},
 {'transform': 'SecondTransform', 'params': {'param3': 'value3'}}]
```

iver56: Looks acceptable to me. What do you think, @mpariente ?

mpariente: This looks ok, specially that Compose will probably have other arguments.

hbredin:
I switched to the `{'transform': ..., 'params': ...}` API even for basic (i.e. not composed) transform.
Bonus: this allows nested composition.

tests/test_config.py
iver56: Yes, let's keep it open until Compose is implemented. I can do that today if I find enough time

iver56 (approved):

iver56:
Thanks!

PR #32, opened 2020-11-05
Implement Compose

### LINKED ISSUE #23: Implement Compose
Like in audiomentations: https://github.com/iver56/audiomentations/blob/master/audiomentations/core/composition.py#L4

iver56:
Close #23

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ from torch_audiomentations import Compose, Gain, PolarityInversion
 
 # Initialize augmentation callable
 apply_augmentation = Compose(
-    [
+    transforms=[
         Gain(
             min_gain_in_db=-15.0,
             max_gain_in_db=5.0,
--- a/README.md
+++ b/README.md
@@ -18,24 +18,30 @@ Audio data augmentation in PyTorch. Inspired by [audiomentations](https://github
 
 ```python
 import torch
-from torch_audiomentations import Gain
+from torch_audiomentations import Compose, Gain, PolarityInversion
 
 
 # Initialize augmentation callable
-apply_gain_augmentation = Gain(
-    min_gain_in_db=-15.0,
-    max_gain_in_db=5.0,
-    p=0.5,
+apply_augmentation = Compose(
+    [
+        Gain(
+            min_gain_in_db=-15.0,
+            max_gain_in_db=5.0,
+            p=0.5,
+        ),
+        PolarityInversion(p=0.5)
+    ]
 )
 
 torch_device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
 
 # Make an example tensor with white noise.
-# This tensor represents 8 audio snippets with 2 channels (stereo) and 2 seconds of 16 kHz audio.
+# This tensor represents 8 audio snippets with 2 channels (stereo) and 2 s of 16 kHz audio.
 audio_samples = torch.rand(size=(8, 2, 32000), dtype=torch.float32, device=torch_device) - 0.5
 
-# Apply gain augmentation. This varies the gain of (some of) the audio snippets in the batch independently.
-perturbed_audio_samples = apply_gain_augmentation(audio_samples, sample_rate=16000)
+# Apply augmentation. This varies the gain and polarity of (some of)
+# the audio snippets in the batch independently.
+perturbed_audio_samples = apply_augmentation(audio_samples, sample_rate=16000)
 ```
 
 # Contribute
--- a/scripts/demo.py
+++ b/scripts/demo.py
@@ -80,7 +80,7 @@ if __name__ == "__main__":
         transforms = [
             {
                 "instance": Compose(
-                    [
+                    transforms=[
                         Gain(
                             min_gain_in_db=-18.0, max_gain_in_db=-16.0, mode=mode, p=1.0
                         ),
@@ -93,7 +93,7 @@ if __name__ == "__main__":
             },
             {
                 "instance": Compose(
-                    [
+                    transforms=[
                         Gain(
                             min_gain_in_db=-18.0, max_gain_in_db=-16.0, mode=mode, p=0.5
                         ),
--- a/scripts/demo.py
+++ b/scripts/demo.py
@@ -8,7 +8,7 @@ import time
 import torch
 from scipy.io import wavfile
 
-from torch_audiomentations import PolarityInversion, Gain, PeakNormalization
+from torch_audiomentations import PolarityInversion, Gain, PeakNormalization, Compose
 
 SAMPLE_RATE = 44100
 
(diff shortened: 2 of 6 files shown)


torch_audiomentations/core/composition.py
```diff
@@ -0,0 +1,62 @@
...
+    ):
+        """
+
+        :param transforms: List of waveform transform instances
+        :param shuffle: Should the order of transforms be shuffled?
+        :param p: The probability of applying the Compose to the given batch.
+        :param p_mode: Only "per_batch" is supported at the moment.
+        """
+        super().__init__()
+        self.transforms = transforms
```
hbredin, line 25:
```suggestion
        self.transforms = torch.nn.ModuleList(transforms)
```
otherwise we'd loose differentiability, wouldn't we?
Maybe add a test in https://github.com/asteroid-team/torch-audiomentations/blob/master/tests/test_differentiable.py?

torch_audiomentations/core/composition.py
```diff
@@ -0,0 +1,62 @@
...
+        self.p = p
+        if p_mode != "per_batch":
+            # TODO: Support per_example as well? And per_channel?
```
hbredin, line 28: I am not sure what would happen when `Compose` is `per_something` and sub-transforms are `per_something_else`.

torch_audiomentations/core/composition.py
```diff
@@ -0,0 +1,62 @@
...
+            raise Exception('p_mode = "{}" is not supported')
+        self.p_mode = p_mode
+        self.shuffle = shuffle
+
+        name_list = []
+        for transform in self.transforms:
+            name_list.append(type(transform).__name__)
+        self.__name__ = "_".join(name_list)
```
hbredin, line 36: Should `__name__` somehow reflect the state of `shuffle`?
Maybe `__name__` is not that important anyway...

For instance, though using emojis is probably not a good idea:
* `Gain ⏩ PolarityInversion` when `shuffle` is `False` 
* `Gain 🔀 PolarityInversion` when `shuffle` is `True`

tests/test_compose.py
```diff
@@ -0,0 +1,106 @@
+import unittest
+
+import numpy as np
+import torch
+from numpy.testing import assert_almost_equal, assert_array_equal
+
+from torch_audiomentations import PolarityInversion, Compose, PeakNormalization
+from torch_audiomentations.augmentations.gain import Gain
```
hbredin, line 8: Why isn't it imported from `torch_audiomentations` like the other ones?

tests/test_compose.py
```diff
@@ -0,0 +1,106 @@
...
+from torch_audiomentations.utils.dsp import convert_decibels_to_amplitude_ratio
+
+
+class TestCompose(unittest.TestCase):
+    def test_compose(self):
+        samples = np.array([[[1.0, 0.5, -0.25, -0.125, 0.0]]], dtype=np.float32)
+        sample_rate = 16000
+
+        augment = Compose(
+            [
```
hbredin, line 18: Nitpicking...
```suggestion
            transforms=[
```

tests/test_compose.py
```diff
@@ -0,0 +1,106 @@
...
+
+            # Either PeakNormalization or Gain was applied last
+            if processed_samples[0, 0, 0] < 0.2:
+                num_gain_last += 1
+            elif processed_samples[0, 0, 0] == 1.0:
+                num_peak_normalization_last += 1
+            else:
+                raise AssertionError("Unexpected value!")
+
+        self.assertGreater(num_peak_normalization_last, 20)
```
hbredin, line 105: Risky but this will hopefully be OK once 2020 is behind us.

README.md
```diff
@@ -18,24 +18,30 @@ Audio data augmentation in PyTorch. Inspired by [audiomentations](https://github
...
+from torch_audiomentations import Compose, Gain, PolarityInversion
 
 
 # Initialize augmentation callable
-apply_gain_augmentation = Gain(
-    min_gain_in_db=-15.0,
-    max_gain_in_db=5.0,
-    p=0.5,
+apply_augmentation = Compose(
+    [
```
hbredin, line 26: Nitpicking...
```suggestion
    transforms=[
```

scripts/demo.py
```diff
@@ -78,6 +78,32 @@ def __exit__(self, type, value, traceback):
     modes = ["per_batch", "per_example", "per_channel"]
     for mode in modes:
         transforms = [
+            {
+                "instance": Compose(
+                    [
```
hbredin, line 83: Nitpicking...
```suggestion
                    transforms=[
```

scripts/demo.py
```diff
@@ -78,6 +78,32 @@ def __exit__(self, type, value, traceback):
...
+                        PeakNormalization(mode=mode, p=1.0),
+                    ],
+                    shuffle=True,
+                ),
+                "name": "Shuffled Compose with Gain and PeakNormalization",
+                "num_runs": 5,
+            },
+            {
+                "instance": Compose(
+                    [
```
hbredin, line 96: Nitpicking...
```suggestion
                    transforms=[
```

torch_audiomentations/core/composition.py
iver56: Let's say you have 8 audio snippets in a batch, and you want to apply the Compose to 50% of them. Then that would be `p=0.5, p_mode="per_example"` (which is not implemented yet). Each transform has its own mode and p_mode that is used independently from other transforms

iver56: I agree that it doesn't make sense to add `mode=` in Compose

torch_audiomentations/core/composition.py
iver56: Yeah, I don't know if this is important. It's copy-pasta from audiomentations 🙈 Should I remove it?

tests/test_compose.py
iver56: Oh, that's probably an error. Let me fix it.

torch_audiomentations/core/composition.py
hbredin: Depends on whether you had any use for this `__name__` thing in mind?

tests/test_compose.py
iver56: Hmm, according to math, this could indeed fail once every ~7 billion times. Should I set a fixed random seed?

https://www.wolframalpha.com/widgets/view.jsp?id=78baf4f3a070cc5b9b226664d2ce80ec

tests/test_compose.py
iver56: Some people will use an arg, some people will use a kwarg :) I added `transforms=` in some of the tests now, so we cover both

torch_audiomentations/core/composition.py
iver56: Nope. I think I'll remove it for now

torch_audiomentations/core/composition.py
iver56: Thanks, nice catch! Done.

torch_audiomentations/core/composition.py
iver56: I left it as a TODO for later: https://github.com/asteroid-team/torch-audiomentations/issues/34

hbredin (approved):

torch_audiomentations/core/composition.py
```diff
@@ -0,0 +1,57 @@
...
+        :param shuffle: Should the order of transforms be shuffled?
+        :param p: The probability of applying the Compose to the given batch.
+        :param p_mode: Only "per_batch" is supported at the moment.
+        """
+        super().__init__()
+        self.transforms = torch.nn.ModuleList(transforms)
+        self.p = p
+        if p_mode != "per_batch":
+            # TODO: Support per_example as well? And per_channel?
+            raise Exception('p_mode = "{}" is not supported')
```
mpariente, line 29:
```suggestion
            raise ValueError(f'p_mode = "{p_mode}" is not supported')
```

iver56: Thanks

PR #50, opened 2020-11-23
Renaming parameters to transform_parameters

### LINKED ISSUE #49: Don't override nn.Module.parameters
We should come up with a different name that doesn't interfere with the nn.Module.parameters method

mogwai:
Closes #49


mpariente:
LGTM

iver56:
Thanks!

PR #51, opened 2020-11-24
Implemented shift/roll batch method

### LINKED ISSUE #41: Improve the execution time of the shift transform?
Like https://github.com/iver56/audiomentations/blob/6f0b8b6c783a6c8268eb00120ab6bb25cab1aab7/audiomentations/augmentations/transforms.py#L273

mogwai:
Sorry for the delay, it took me a while to figure out how to get the shift without roll working. I've included some randomness in the method itself but noticed that we should be deriving the values to roll from the transform parameters. What are yours thoughts on this implementation?

I'll need to fiddle with the tests a tad I think too. 

Closes #41


iver56:
> I've included some randomness in the method itself but noticed that we should be deriving the values to roll from the transform parameters. What are yours thoughts on this implementation?

I would prefer keeping all randomness in randomize_parameters. These are the main motivations for doing randomization in a separate step:

1) So we can call `freeze_parameters` and apply the exact same transform to several inputs. E.g. transform a target sound in the same way as the input sound.
2) So we can extract the applied parameters and store them for e.g. inspection/visualization later

torch_audiomentations/augmentations/shift.py
```diff
@@ -71,8 +71,10 @@ def randomize_parameters(
             <= max_shift_in_samples
             <= torch.iinfo(torch.int32).max
         )
-
-        selected_batch_size = selected_samples.size(0)
+        if selected_samples.dim() == 2:
+            selected_batch_size = 1
```
iver56, line 75: If the tensor is two-dimensional, it it assumed that the audio inside it is mono and the first dimension is the batch size.
This means that we should do `selected_batch_size = selected_samples.size(0)` even when the tensor is two-dimensional

iver56:
Cool 👍 I want to do some benchmarking to find out how much faster this approach is

torch_audiomentations/augmentations/shift.py
mogwai: Ah ok I was testing on multi channel 2d samples. I think it would be generally better to always expect a `b c t` shape. I think you said that somewhere too?

iver56: Yeah, that does makes sense. It's very small price to pay for those who do only mono audio, not multichannel.

mogwai: What would you like me to change for this PR then?

iver56: You can let `Shift` assume that the input tensor is always 3-dimensional

torch_audiomentations/augmentations/shift.py
```diff
@@ -90,20 +92,39 @@ def randomize_parameters(
...
-                elif num_samples_to_shift < 0:
-                    selected_samples[i, ..., num_samples_to_shift:] = 0.0
-
-        return selected_samples
+        """
+        b, c, t = tensor.shape
+        # Max to roll by
+        
+        # Arange indexes
+        x = torch.arange(t)[None, None, :].repeat(b, c, 1)
```
iver56, line 109:
```suggestion
        x = torch.arange(t, device=tensor.device)[None, None, :].repeat(b, c, 1)
```

torch_audiomentations/augmentations/shift.py
```diff
@@ -90,20 +92,39 @@ def randomize_parameters(
...
+        
+    
+        # Apply Roll
+        r = r[:, None, None]
+        idxs = (x - r)
+        
+        # Back to flattened indexes
+        add = (torch.arange(b) * t * c)[:,None,None].repeat(1, c, t)
```
iver56, line 117:
```suggestion
        add = (torch.arange(b, device=tensor.device) * t * c)[:,None,None].repeat(1, c, t)
```

iver56 (changes requested):

iver56:
I ran my execution time benchmark script (from https://github.com/asteroid-team/torch-audiomentations/pull/52) on this, and so far it seems that the naive approach with the for-loop is generally a bit faster:

device_name_plot
num_samples_plot
num_channels_plot
batch_size_plot

I'll run this once more where I uncomment @torch.jit.script

iver56:
I get an error like this when I uncomment @torch.jit.script:

```
Traceback (most recent call last):
  File "C:/Users/Iver/Code/torch-audiomentations/scripts/perf.py", line 129, in <module>
    device,
  File "C:/Users/Iver/Code/torch-audiomentations/scripts/perf.py", line 76, in measure_execution_time
    transform(samples=samples, sample_rate=sample_rate).cpu()
  File "C:\Users\Iver\Anaconda3\envs\torch-audiomentations-gpu\lib\site-packages\torch\nn\modules\module.py", line 722, in _call_impl
    result = self.forward(*input, **kwargs)
  File "C:\Users\Iver\Code\torch-audiomentations\torch_audiomentations\core\transforms_interface.py", line 157, in forward
    ] = self.apply_transform(selected_samples, sample_rate)
  File "C:\Users\Iver\Code\torch-audiomentations\torch_audiomentations\augmentations\shift.py", line 98, in apply_transform
    return self.shift(selected_samples, r, self.rollover)
RuntimeError: shift() Expected a value of type 'Tensor' for argument 'r' but instead found type 'bool'.
Position: 2
Value: True
Declaration: shift(Tensor self, Tensor tensor, Tensor r, bool rolling=False) -> (Tensor)
Cast error details: Unable to cast Python instance to C++ type (compile in debug mode for details)
```

iver56:
The jit script error was related to the "self" argument. I was able to run the same benchmark with torch.jit.script enabled on Shift (but not ForLoopShift). It still seems a bit slower:

device_name_plot
num_samples_plot
num_channels_plot
batch_size_plot

I also did some a little memory testing, and it seems like this alternative Shift implementation takes roughly 4 times more memory.

I really do appreciate the effort that you put into this, but I'm afraid that I'm going to have to close this pull request since it turns out that the approach with the for loop and torch.roll is faster and less memory-hungry

mogwai:
Could you share the code you used to benchmark?

mogwai:
These benchmark graphs are really good by the way nice work :) I made some changes after looking into some neater ways of doing things

I noticed that you weren't torch.cuda.synchronize() in the perf_benchmarks code. This is bad for my version of shift.

```
params = dict(
        batch_sizes=[2, 8, 16, 32, 64, 128, 256],
        channels=[1, 2, 3],
        durations=[1, 2, 4],
        sample_rates=[8000, 16000],
        devices=["cpu", "cuda"],
    )
```

image
image

I also thought it would be interesting to look at the graphs purely on cuda tensors as this is where this library is hoping to shine?
image
image
image

iver56:
I like that you don't give up so easily :)

Re torch.cuda.synchronize(): I didn't think it was needed, because I did a little `.cpu()` at the end, which forces the script to wait until CUDA is done. If it matters, I stand corrected.

If the new Shift implementation (with your refinements) is significantly faster on CUDA, while the old one is significantly faster on CPU, I guess we could automatically choose the fastest implementation based on the nature of the data that comes in. I think albumentations does a lot of that too.

mogwai:
I was interested in how you did the memory profile too. Could that be something that is added to the profiling code you've added. I'm assuming using expand over repeat has made a different to the memory consumption but wasn't sure how to test it invidually for each method. I'm also taking advantage of broadcasting a bit better. 

The `.cpu()` probably does do that actually so either would work. 

I was also going to suggest that we choose the best implementation at the time as well. Sounds like a good idea until someone can come along and think of a better way of doing them even faster. There must be a magic torch method I'm missing here. 

Comparing with `rolling=True` as well think that the torch.gather method was faster on the cpu. 

Another idea I thought about was using 1D Convolutions.

iver56:
The memory "profile" I did was not very scientific or accurate, but it was more about checking which one used the least amount of memory. I tried to see how many seconds of audio I could pass through the transform without getting an out-of-memory error on a GPU with 4 GB RAM. This way I found out that the old shift could do up to around 16 * 8 * 480 seconds while the alternative could do up to around 16 * 8 * 120 seconds.

iver56:
If you know of a way of doing proper memory profiling, then that would be nice to have in the benchmark code indeed! 👍

mogwai:
This is ready for another review.

Adding more code reduced the coverage requirements though.

iver56:
Thanks!

iver56 (approved):

PR #59, opened 2020-12-04
Compose compatible with non audiomentation transforms

mogwai:
I was trying to Compose transforms that don't have a sample rate defined in the method declaration and ran into an error.

### DIFF
--- a/torch_audiomentations/core/composition.py
+++ b/torch_audiomentations/core/composition.py
@@ -4,6 +4,8 @@ from typing import List
 import torch
 import typing
 
+from torch_audiomentations.core.transforms_interface import BaseWaveformTransform
+
 
 class Compose(torch.nn.Module):
     """This class can apply a sequence of transforms to waveforms."""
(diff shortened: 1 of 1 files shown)


iver56:
Thanks for the fix!

torch_audiomentations/core/composition.py
```diff
@@ -37,8 +39,13 @@ def forward(self, samples, sample_rate: typing.Optional[int] = None):
...
                 random.shuffle(transform_indexes)
             for i in transform_indexes:
-                samples = self.transforms[i](samples, sample_rate)
+                tfm = self.transforms[i]
+                if isinstance(tfm, BaseWaveformTransform):
+                    samples = self.transforms[i](samples, sample_rate)
+                else:
+                    samples = self.transforms[i](samples)
 
+                samples = self.transforms[i](samples, sample_rate)
```
iver56, line 48: This line needs to be removed. The two tests that are currently failing will pass again after the removal

iver56 (changes requested):

torch_audiomentations/core/composition.py
mogwai: Yes wow, should have seen that!

PR #69, opened 2020-12-15
When looking for audio files, follow symlinks by default.

iver56:
Relates to https://github.com/asteroid-team/torch-audiomentations/commit/01f1436481c3b8f67a30caf7e0578a183044ef0f

@hbredin

### DIFF
--- a/.github/workflows/test_requirements.txt
+++ b/.github/workflows/test_requirements.txt
@@ -1,6 +1,6 @@
 numpy>=1.14,<1.18
 scipy>=1.5.2
-torch>=1.6.0
+torch==1.7.0
 torchaudio>=0.6.0
 audiomentations>=0.11.0
 audioread>=2.1.8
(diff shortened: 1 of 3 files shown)


.github/workflows/test_requirements.txt
```diff
@@ -1,6 +1,6 @@
 numpy>=1.14,<1.18
 scipy>=1.5.2
-torch>=1.6.0
+torch==1.7.0
```
hbredin, line 3: Was this change intended?

hbredin (approved):
LGTM. Would have enjoyed a few comments in the unit test though 👍

.github/workflows/test_requirements.txt
iver56: Not sure. But the symptom was that the CI did not successfully run. I think it was due to torch 1.7.1 (the 700 MB variant) being download instead of the 1.7.0 CPU one (~60 MB), which we ran in tests earlier.

Issue #46, opened 2020-11-17
Implement white/brown/pink/gaussian noise


iver56:

amiasato:
Hi, I'm willing to contribute to the project.

Aside from white noise, all other types are easier to implement in spectral space. Since other augmentations (e.g. pitch shift) should benefit from spectral manipulation, i guess something like the [BaseSpectrogramTransform](https://github.com/iver56/audiomentations/blob/master/audiomentations/core/transforms_interface.py#L84) could be implemented. However, the mentioned interface works only with magnitudes, whilst I believe the best course would be to use the  complex spectrum which allows for perfect signal reconstruction. Do you have any thoughts on these issues?

iver56:
Hi amiasato. Cool that you want to contribute! I'm looking forward to it.

Yes :) BaseSpectrogramTransform is a welcome addition. If we consider magnitude spectrograms and complex spectrograms two different beasts, maybe we can call them BaseMagnitudeSpectrogramTransform and BaseComplexSpectrogramTransform? Thinking of it, most spectrogram transforms should ideally support both magnitude and complex input. Some transforms (like frequency mask à la SpecAugment) would leave the phase of any complex spectrogram untouched though. So maybe we should have a single base class, like BaseSpectrogramTransform, and each transform can have a flag like `supports_complex_spectrograms` and `supports_magnitude_spectrograms`? And it could check the format of the input to determine whether it's complex or not?

amiasato:
The only problem I see with the auto-inference of the input type is that we can't actually know if a complex input is in the linear scale (real/imag) or in the log scale (mag/phase). Torchaudio's "[Spectrogram](https://pytorch.org/audio/stable/transforms.html#spectrogram)" transform has an API which returns linear scale when "power" is None, so I guess we may assume the former most of the time. Using two different classes seem like overkill to me, the flags alternative indeed sounds nicer.

In any case, I am still wrapping my head around the multiple torch APIs for short-time transforms ([stft](https://pytorch.org/docs/stable/generated/torch.stft.html) and [istft](https://pytorch.org/docs/stable/generated/torch.istft.html) are alternatives), so for starters I guess I will use those for this issue in particular.

PR #80, opened 2021-04-09
Colored noise

amiasato:
Addresses #46. API is basically derived from BG noise, switching the background list by the `f_decay` parameter which defines the "color" of the noise, and is randomized. Tests were also mostly copied from BG noise, with the only addition of invalid parameter tests for the `f_power` range.

In the end, the colored noise implementation was done with plain RFFTs over one-second gaussian noise which is replicated until `num_samples`, nothing too fancy. Gray noise was left out because it has a custom "constant loudness" envelope which I couldn't think of a proper way to randomize as a parameter, so maybe it can be implemented as its own transform.

### DIFF
--- a/tests/test_colored_noise.py
+++ b/tests/test_colored_noise.py
@@ -1,0 +1,81 @@
+import random
+import unittest
+
+import torch
+
+from torch_audiomentations import AddColoredNoise
+from torch_audiomentations.utils.io import Audio
+from .utils import TEST_FIXTURES_DIR
+
+
+class TestAddBackgroundNoise(unittest.TestCase):
+    def setUp(self):
+        self.sample_rate = 16000
+        self.audio = Audio(sample_rate=self.sample_rate)
+        self.batch_size = 16
+        self.empty_input_audio = torch.empty(0)
+
+        self.input_audio = self.audio(
+            TEST_FIXTURES_DIR / "acoustic_guitar_0.wav"
+        ).unsqueeze(0)
+
+        self.input_audios = torch.cat([self.input_audio] * self.batch_size, dim=0)
+        self.cl_noise_transform_guaranteed = AddColoredNoise(20, p=1.0)
+        self.cl_noise_transform_no_guarantee = AddColoredNoise(20, p=0.0)
+
+    def test_colored_noise_no_guarantee_with_single_tensor(self):
+        mixed_input = self.cl_noise_transform_no_guarantee(
+            self.input_audio, self.sample_rate
+        )
+        self.assertTrue(torch.equal(mixed_input, self.input_audio))
+        self.assertEqual(mixed_input.size(0), self.input_audio.size(0))
+
+    def test_background_noise_no_guarantee_with_empty_tensor(self):
+        with self.assertWarns(UserWarning) as warning_context_manager:
+            mixed_input = self.cl_noise_transform_no_guarantee(
+                self.empty_input_audio, self.sample_rate
+            )
+
+        self.assertIn(
+            "An empty samples tensor was passed", str(warning_context_manager.warning)
+        )
+
+        self.assertTrue(torch.equal(mixed_input, self.empty_input_audio))
+        self.assertEqual(mixed_input.size(0), self.empty_input_audio.size(0))
+
+    def test_colored_noise_guaranteed_with_zero_length_samples(self):
+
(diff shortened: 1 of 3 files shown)


torch_audiomentations/augmentations/colored_noise.py
```diff
@@ -0,0 +1,117 @@
...
+from ..core.transforms_interface import BaseWaveformTransform
+from ..utils.dsp import calculate_rms
+from ..utils.io import Audio
+
+
+def _gen_noise(f_decay, num_samples, sample_rate):
+    """
+    Generate colored noise with f_decay decay using torch.fft
+    """
+    noise = torch.normal(0, 1, (sample_rate,))
```
iver56, line 14: Could we ensure that this tensor is created on the same device as the input audio? That way we can avoid moving data from the CPU to the GPU

Something like this? https://stackoverflow.com/a/59179927/2319697

tests/test_colored_noise.py
```diff
@@ -0,0 +1,81 @@
...
+import unittest
+
+import torch
+
+from torch_audiomentations import AddColoredNoise
+from torch_audiomentations.utils.io import Audio
+from .utils import TEST_FIXTURES_DIR
+
+
+class TestAddBackgroundNoise(unittest.TestCase):
```
iver56, line 11: This should ideally be renamed to TestAddColoredNoise

iver56 (changes requested):
Thanks for the contribution!

How fast is this approach that involves rfft and irfft? It would be interesting to have it added to the list of transforms to run in the perf benchmark:
https://github.com/asteroid-team/torch-audiomentations/blob/master/scripts/perf_benchmark.py#L117

I can imagine that e.g. white noise can be generated much faster since it doesn't need rfft+irfft

torch_audiomentations/augmentations/colored_noise.py
```diff
@@ -0,0 +1,117 @@
...
+    return noise[:num_samples]
+
+
+class AddColoredNoise(BaseWaveformTransform):
+    """
+    Add colored noises to the input audio.
+    """
+
+    supports_multichannel = True
+    requires_sample_rate = True
```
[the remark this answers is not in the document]
amiasato: The sample rate is used as the "f" parameter of the frequency decay envelope  (1/f^f_power).

torch_audiomentations/augmentations/colored_noise.py
amiasato: Seems feasible enough, I will adjust accordingly.

tests/test_colored_noise.py
amiasato: Sloppiness of mine, I'll change accordingly

amiasato:
Regarding white noise, I initially intended to have static generators for each colored noise with a fixed f_decay, and an f_decay==0 would bypass the FFT. I could create a separate white noise generator, or even a "colorless" noise transform which would sample between white and gray noise. In any case, I will  benchmark white vs color noise with the aforementioned script and return when done.

amiasato:
num_samples_plot
device_name_plot
num_channels_plot
batch_size_plot

Env:
CPU: 6-Core Intel Core i5-8400
GPU: GeForce GTX 1060 6GB
torch==1.8.1; torchaudio==0.8.1; Cuda compilation tools, release 10.1, V10.1.243 (no cuDNN)

I used two implementations of white noise - `AddWhiteNoise` draws `num_samples` from a normal distribution, while `AddWhiteNoise1` draws 8000 samples and concatenates (regardless of `sample_rate`, which is not needed for white noise). I compared them with the implemented colored noise, polarity inversion and impulse response transforms. As expected, the higher the `num_samples`, the lesser the gap between the white and colored noises, as the latter only computes a fixed-size FFT regardless of the input length. The same applies for `num_channels`.

Given the time differences, is adding `WhiteNoiseTransform` desirable/justified?

iver56:
Thanks for making the chart! It looks like WhiteNoiseTransform is roughly 4-5 times faster. It can be added later. For now, let's merge AddColoredNoise.

Btw, I noticed that this transform uses torch.fft.rfft, which was introduced in torch 1.7.0. Should we use the "polyfill" that is `from torch_audiomentations.utils.fft import rfft, irfft` here, or increase the minimum supported torch version?

iver56 (approved):

PR #82, opened 2021-06-18
Add PitchShift operation

KentoNishi:
This PR is still a work in progress, but here is the gist of it:

* I made a separate library called [torch-pitch-shift](https://pypi.org/project/torch-pitch-shift/) that calculates pitch-shift targets that can be computed quickly using tensor operations beforehand. The `PitchShift` class basically just calls this library to transform each sample in the batch. 
* One caveat is that the current implementation requires the sample rate to be passed during the `Compose` process. I'm sure yall can figure something out, the current class is just a proof of concept.
* I haven't added any tests. Please help! (the only demo file present is `test.py` at the root of the repo, which I'll have to delete later.)

The library I made is still undocumented, so I don't want to make it public just yet. If you want to verify that the code is not malicious, let me know so I can add you to the repo!

### DIFF
--- a/tests/test_impulse_response.py
+++ b/tests/test_impulse_response.py
@@ -54,10 +54,19 @@ def test_impulse_response_guaranteed_with_single_tensor_input(ir_transform, inpu
     assert not torch.equal(mixed_input, input_audio)
 
 
+@pytest.mark.parametrize(
+    "compensate_for_propagation_delay",
+    [False, True],
+)
 def test_impulse_response_guaranteed_with_batched_tensor_input(
-    ir_transform, input_audios
+    ir_path, sample_rate, input_audios, compensate_for_propagation_delay
 ):
-    mixed_inputs = ir_transform(input_audios)
+    mixed_inputs = ApplyImpulseResponse(
+        ir_path,
+        compensate_for_propagation_delay=compensate_for_propagation_delay,
+        p=1.0,
+        sample_rate=sample_rate,
+    )(input_audios)
     assert mixed_inputs.shape == input_audios.shape
     assert not torch.equal(mixed_inputs, input_audios)
 


KentoNishi:
It's in a working and semi-usable state!

## Input
```python
# TODO: WRITE SOME REAL TESTS

import torch
from torch_audiomentations import Compose, PitchShift


# Initialize augmentation callable
apply_augmentation = Compose(
    transforms=[
        PitchShift(16000, p=1),
    ]
)

torch_device = torch.device("cuda" if torch.cuda.is_available() else "cpu")

# Make an example tensor with white noise.
# This tensor represents 8 audio snippets with 2 channels (stereo) and 2 s of 16 kHz audio.
import time

audio_samples = (
    torch.rand(size=(8, 2, 32000), dtype=torch.float32, device=torch_device) - 0.5
)

# Apply augmentation. This varies the gain and polarity of (some of)
# the audio snippets in the batch independently.
start = time.process_time()
perturbed_audio_samples = apply_augmentation(audio_samples, sample_rate=16000)
print(time.process_time() - start)
```
## Output
```
1.53125
```

KentoNishi:
i haven't implemented any tests, I'll leave that up to you guys :) 

plz let me know what you think of what I have so far!

iver56:
Thanks for making this, and thank you for your patience. I'll have a look at this soon 👍

KentoNishi:
i tried using it in my own project, and it seems like there's a memory leak somewhere? vram usage keeps increasing when i include pitch shift. will investigate.

iver56:
Interesting. The memory leak should ideally be fixed before we merge this :)

test_pitch_shift.py
```diff
@@ -0,0 +1,28 @@
+# TODO: WRITE SOME REAL TESTS
+
+import torch
+from torch_audiomentations import Compose, PitchShift
+
+
+# Initialize augmentation callable
+apply_augmentation = Compose(
```
iver56, line 8: It would be nice if you could wrap this code in a test function, so the test runnner doesn't automatically run this code while importing test_*.py files, looking for test functions.

torch_audiomentations/augmentations/pitch_shift.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+
+    supports_multichannel = True
+    requires_sample_rate = True
+
+    def __init__(
+        self,
+        sample_rate: int,
+        min_transpose_ratio=0.5,
+        max_transpose_ratio=2,
```
iver56, line 21:
```suggestion
        max_transpose_ratio: float = 2.0,
```

torch_audiomentations/augmentations/pitch_shift.py
```diff
@@ -0,0 +1,73 @@
...
+        :param sample_rate:
+        """
+        super().__init__(mode, p, p_mode, sample_rate)
+
+        self._min_transpose_ratio = min_transpose_ratio
+        self._max_transpose_ratio = max_transpose_ratio
+        if self._min_transpose_ratio > self._max_transpose_ratio:
+            raise ValueError("max_transpose_ratio must be > min_transpose_ratio")
+        self._pitch_shift = PitchShifter()
+        self._sample_rate = sample_rate
```
iver56, line 41: At this point we need to assert that a valid (not None) sample rate is specified

torch_audiomentations/augmentations/pitch_shift.py
```diff
@@ -0,0 +1,73 @@
...
+        batch_size, _, num_samples = selected_samples.shape
+
+        self.transform_parameters["transpositions"] = choices(
+            self._fast_shifts, k=batch_size
+        )
+
+    def apply_transform(self, selected_samples: torch.Tensor, sample_rate: int = None):
+        batch_size, num_channels, num_samples = selected_samples.shape
+
+        if sample_rate is None:
```
iver56, line 62: If the sample_rate param here is not None, check if there is a mismatch between the sample rate given here and the one given when initing PitchShift. If there is a mismatch, raise an exception explaining what went wrong, and how to fix it

torch_audiomentations/augmentations/pitch_shift.py
```diff
@@ -0,0 +1,73 @@
+from random import choices
+import torch
+import torch.nn.functional as F
+
+from ..core.transforms_interface import BaseWaveformTransform
+from torch_pitch_shift import *
```
iver56, line 6: I would prefer to be explicit about what is imported here

E.g.
`from torch_pitch_shift import get_fast_shifts, PitchShifter`

or something like that

iver56 (changes requested):

KentoNishi:
Thanks for the reviews, will

torch_audiomentations/augmentations/pitch_shift.py
```diff
@@ -0,0 +1,72 @@
...
+            self._fast_shifts, k=batch_size
+        )
+
+    def apply_transform(self, selected_samples: torch.Tensor, sample_rate: int = None):
+        batch_size, num_channels, num_samples = selected_samples.shape
+
+        if sample_rate is None:
+            sample_rate = self._sample_rate
+
+        for i in range(batch_size):
```
iver56, line 65: Is there a way to make the computation "batched", so we can get by without this for loop?

KentoNishi: The shift `__call__` looks like this:

```python
    def __call__(
        self, input: torch.Tensor, shift: Union[float, Fraction], sample_rate: int
    ) -> torch.Tensor:
        """
        Shift the pitch of a waveform by a given amount.

        Parameters
        ----------
        input: torch.Tensor [shape=(channels, samples)]
            Input audio clip of shape (channels, samples)
        shift: float OR Fraction
            `float`: Amount to pitch-shift in # of bins. (1 bin == 1 semitone if `bins_per_octave` == 12)
            `Fraction`: A `fractions.Fraction` object indicating the shift ratio. Usually an element in `get_fast_shifts()`.
        sample_rate: int
            The sample rate of the input audio clip.

        sample_rate: int

        Returns
        -------
        output: torch.Tensor [shape=(channels, samples)]
            The pitch-shifted audio clip
        """
        if not isinstance(shift, Fraction):
            shift = 2.0 ** (float(shift) / self._bins_per_octave)
        resampler = T.Resample(sample_rate, int(sample_rate / shift)).to(input.device)
        output = input
        output = resampler(output)
        output = torch.stft(output, self._n_fft)[None, ...]
        stretcher = T.TimeStretch(
            fixed_rate=float(1 / shift), n_freq=output.shape[2]
        ).to(input.device)
        output = stretcher(output)
        output = torch.istft(output[0], self._n_fft)
        del resampler, stretcher
        if output.shape[1] >= input.shape[1]:
            return output[:, : (input.shape[1])]
        return pad(output, pad=(0, input.shape[1] - output.shape[1], 0, 0))
```

From what I know, there isn't a way to apply variable TimeStretch or Resample operations at variable on a batch. I guess we can add a mode to apply the same random augmentation to each batch so they can be done in parallel?

KentoNishi: To clarify, if we do add a mode that allows for all samples in one batch to be randomly shifted by the same factor, it would basically be the same as creating one track with a large number of channels. That could definitely work

KentoNishi:
Alright gonna get some sleep now, good night 😴

torch_audiomentations/augmentations/pitch_shift.py
iver56: Ok, I see. We could consider it future work then.

KentoNishi:
Adding `Batched_Pitch_Shift` as a separate class might be a good idea. Currently working on batched shifting in the library itself :)

KentoNishi:
Implementation of batches transforms is done in the library! Will update the fork when I have time later.

KentoNishi:
image

timed in seconds. really liking how fast it runs! this is with 8 samples with sr=16000 and clips each 2 seconds long

KentoNishi:
@iver56 what's new:

* wrote tests, they pass now :)
* added 3 modes: `per_batch`, `per_example`, and `per_channel`. `per_batch` is default cuz it's fast
* added assertions and error messages

iver56:
Thanks for the improvements 👍 I will re-review this soon-ish (my availability is a bit limited, thanks for your patience)

iver56:
I would like "per_example" to be the default mode. Although "per_batch" is faster, variation within each batch is typically a good idea when training models :)

iver56:
The other transforms have "per_example" as the default mode too

iver56:
Something happened to the commits after the force push - I can't see the commits in the pull request anymore

iver56:
I think would also prefer a bit more modest default parameters - pitch shifting a whole octave up or down is a bit extreme. In audiomentations the default is -4 to +4 semitones. -4 semitones is down a third of an octave, and +4 semitones is up a third of an octave. This default would give a range of two thirds of an octave.

In audiomentations, the pitch shifting parameters are input as semitones. Could that be relevant here too? I personally find it easier to relate to the numbers when they are given in semitones (e.g. -12 and +12) instead of fractions (e.g. 0.5 and 2.0)

KentoNishi:
@iver56 I think what happened is that the force push overwrote all my commits. I'll patch it up and open a new PR

Issue #60, opened 2020-12-07
Optionally compensate for propagation delay in ApplyImpulseResponse


iver56:
Maybe we could use this? https://github.com/jonashaag/numpy-fast-align-mse

mpariente:
This would be cool but it's on CPU, without batch.. 
Argmax of the abs on the RIR gives the delay, would that be ok?

iver56:
Yes, in most cases that works well enough

jonashaag:
FYI I use argmax for RIR delay and the MSE brute force method for other delays (eg. ffmpeg introduces random delays on left of a signal if you encode with some particular codecs)

hbredin:
I just opened a PR that does just that (argmax of abs): #83

PR #83, opened 2021-06-21
feat: add fix_propagation_delay option

hbredin:
Addresses #60


hbredin:
I have no strong opinion on the name of this option. Feel free to choose a better name :)

iver56:
Thanks for the contribution, and thanks for the patience 😄 I will review this soon-ish (my availability is a bit limited)

torch_audiomentations/augmentations/impulse_response.py
```diff
@@ -74,11 +77,29 @@ def apply_transform(self, selected_samples, sample_rate: int = None):
...
+                        :, propagation_delay : propagation_delay + num_samples
+                    ]
+                    for convolved_sample, propagation_delay in zip(
+                        convolved_samples, propagation_delays
+                    )
+                ],
+                dim=0,
+            )
+
+            return convolved_samples
```
iver56, line 102: Why don't we want to do `[..., :num_samples]` here? In my code, I did it to ensure that the output has the same shape as the input, instead of being longer. In other words, I cut off the reverb tail

hbredin: Line 92 can be written as:

```python
convolved_sample[:, propagation_delay:][...,:num_samples]
```

1. We start by getting rid of the first `propagation_delay` with `[..., propagation_delay:]` 
2. We cut off the reverb tail as you suggest with `[..., :num_samples]`

Point 1. is what the `fix_propagation_delay` option is used for.
Or maybe I completely got wrong this propagation delay thing  (that is definitely an option :))

iver56: Ah, you're right - I didn't read the slice code above in detail. This works as expected already.

hbredin:
Thanks for merging!

iver56:
You're welcome :) Again, thanks for the contribution.

I did some minor changes after merging:
* Test coverage
* Renamed the parameter and added an explanation in the docstring

PR #85, opened 2021-06-28
Add PitchShift augmentation

KentoNishi:
The original PR (#82) was messed up by a force push. I've recovered the lost commits and merged upstream changes


KentoNishi:
Comments to resolve: 

[ ] I would like "per_example" to be the default mode. Although "per_batch" is faster, variation within each batch is typically a good idea when training models :)
[ ] I think would also prefer a bit more modest default parameters - pitch shifting a whole octave up or down is a bit extreme. In audiomentations the default is -4 to +4 semitones. -4 semitones is down a third of an octave, and +4 semitones is up a third of an octave. This default would give a range of two thirds of an octave. In audiomentations, the pitch shifting parameters are input as semitones. Could that be relevant here too? I personally find it easier to relate to the numbers when they are given in semitones (e.g. -12 and +12) instead of fractions (e.g. 0.5 and 2.0)

iver56:
I tried to install this now. It runs, but the result sounds oddly distorted. Is this as intended, or is something wrong on my end?

I have attached one of the results I got in torch-audiomentations along with a reference (in mono) made with the pitch shift in audiomentations:

[pitch-shift-results.zip](https://github.com/asteroid-team/torch-audiomentations/files/6733392/pitch-shift-results.zip)

amiasato:
If you guys don't mind (especially @KentoNishi), I have a not-entirely-naive-but-almost pitch shift implementation which I was fiddling with a couple of weeks ago, but the whole "per batch" X "per example" interfaces kinda hindered my PR plans. Since as it stands, this part was sorted out by this PR, I could either take on from here or you could check this [gist](https://gist.github.com/amiasato/535affc335cdc00f11e83d8bf994a2fb) before deciding which implementation to choose.

KentoNishi:
Thanks for the heads up @amiasato! I was able to implement batched transforms so I'll go with my implementation. Code looks really similar though.

@iver56 i think the distortion issue is caused by a low n_fft (I'm using 256 by default). I'll see what I can do to raise the quality, sorry I haven't had much time to get back to work on it 😔

iver56:
@amiasato Sorry if the "per batch"/example/channel thing sounded like a hindrance. In fact, I have implemented that logic in the base class, so the transform implementation actually doesn't have to do much special handling of those three cases :) The concept is that the base class will reshape the 3D tensor according to the chosen mode, and the transform (subclass) will always apply an effect with independent parameters for each example in the batch.

amiasato:
@iver56 np at all, I just didn't have to deal too much with it in the first PR because I mostly copy-pasted from the additive noise aug. However, in this case, having separate per-batch and per-example implementations is imperative due to performance discrepancies, so I would need to spend more time than originally planned on it.

@KentoNishi glad to hear that, with pitch shift, at least for my needs the "bread and butter" of audio augmentations should be covered, so I am really looking forward to this PR. Thanks for the feature!

KentoNishi:
Looks like I gotta stop slackin and get back to torchin huh 😆

iver56:
> @iver56 i think the distortion issue is caused by a low n_fft (I'm using 256 by default). I'll see what I can do to raise the quality, sorry I haven't had much time to get back to work on it 😔

Maybe it makes sense to have a n_fft value that somehow depends on the sample rate? E.g. 44,1 kHz typically requires a larger n_fft than 16 kHz. In other words, if a given value of n_fft sounds fine when applied to 16 kHz audio, that same n_fft value may sound bad if the audio has a high sample rate.

KentoNishi:
@amiasato I took inspiration from your //64 "rule of thumb". That's the default behavior now

KentoNishi:
I made the library repository public as well: https://github.com/KentoNishi/torch-pitch-shift

KentoNishi:
https://user-images.githubusercontent.com/38841491/124249028-504dd000-dad8-11eb-9778-4c20a21aff07.mp4

KentoNishi:
@iver56 just a ping to remind you to look at this pr again when you have time 😊

iver56:
Thanks for the reminder, and thanks for the patience! I am currently on a trip without a laptop available, but I will get to this when I'm back from the vacation 👍Expect at least a week

iver56:
I gave this a spin, and it sounds a lot better now. However, it still sounds a bit distorted on the 44,1 kHz audio clip I gave it. I've pasted some spectrograms below for reference. You can see in the spectrogram how the transform introduces high-frequency harmonics. Not sure what needs to be done to fix this.

bilde

iver56:
[PitchShift_per_batch_000_perfect-alley1.zip](https://github.com/asteroid-team/torch-audiomentations/files/6839122/PitchShift_per_batch_000_perfect-alley1.zip)

KentoNishi:
Currently I'm resampling first, time stretching second. Pretty sure librosa does it the other way. Maybe I should try that?

iver56:
It's worth a try

Btw, if you want to reproduce my run, you could insert this in scripts/demo.py:

```
            {
                "get_instance": lambda: PitchShift(
                    sample_rate=SAMPLE_RATE, mode=mode, p=1.0
                ),
                "num_runs": 1,
            },
```

and then run `python -m scripts.demo` from the torch-audiomentations root directory.

You could also include this change in the pull request for future pitch shift tinkerers.

iver56:
Something happened with the pull request. A bot force-pushed to your branch?

KentoNishi:
ahah, looks like it's time to force push from my local XD

PR #92, opened 2021-07-22
draft band pass filter and its test

keunwoochoi:
It's a minimal edit from `LowPassFilter`.


iver56:
Thanks for the contribution! BPF is certainly a welcome addition.

Here's some feedback:

I think it's better to randomize center frequency and bandwidth (and compute low cutoff and high cutoff from that).

The pull request's current approach can be a bit problematic if a user wants a bandpass augmentation that can go almost all over the place, e.g.

low_cutoff_freq_range: tuple = (20, 6000),
high_cutoff_freq_range: tuple = (1000, 19000),

Would often give a high cutoff that is lower than the low cutoff, which results in all audio being "accidentally" filtered away (in theory). Controlling a center frequency and a bandwidth ensures that the high cutoff is always greater than the low cutoff.

I do realize that controlling low cutoff and high cutoff independently can be useful in some cases, but in that case I think the user is better off with a Compose that contains a LPF followed by a HPF

iver56:
Btw, this sparked an idea: We should have a band reject (AKA band stop filter) transform as well :)

keunwoochoi:
> I do realize that controlling low cutoff and high cutoff independently can be useful in some cases, but in that case I think the user is better off with a Compose that contains a LPF followed by a HPF

That's what I wanted to do. Is there a way to let them rely on the same randomization? Otherwise, it could be one of {LPF, BPF, HPF} if p < 1.0.

iver56:
> Is there a way to let them rely on the same randomization? Otherwise, it could be one of {LPF, BPF, HPF} if p < 1.0.

Hmm, interesting question. I think the answer is no right now. Maybe we should make a wrapper class that takes a parameter that is a transform. It could do the following:

1) random selection
2) slice and reshape the tensor
3) apply the given transform (it could be a Compose)
4) reshape back, include any untouched data again, and return the augmented tensor

similar to what the base class does right now

Then we could have code like

```
RandomSelection(
    Compose([LowPassFilter(p=1.0), HighPassfilter(p=1.0)]),
    p_mode="per_example",
    p=0.5
)
```

Making a data augmentation lib that caters everyone is hard :)

keunwoochoi:
it certainly is :)

i agree that defining f_c and bandwidth would be more suitable. i'll make that change and push soon.

iver56:
Thanks

keunwoochoi:
any recommendations for the names and the default values of the ranges of center frequency and bandwidth?

iver56:
Hmm, maybe something like this:

```
min_center_frequency=200,
max_center_frequency=7500,
min_bandwidth_fraction=0.25,
max_bandwidth_fraction=2.0
```

where the bandwidth is a fraction of the center frequency.

I don't know exactly how this will sound, but once we put this transform in scripts/demo.py and run it, we could listen to some audio outputs and tweak the default params accordingly :)

keunwoochoi:
Okay, I'll use `_fraction` since it's probably more accessible to a wider range of users.

iver56:
According to the CI, one test fails:

```
>       self.transform_parameters["bandwidth"] = torch.distribution.Uniform(
            low=self.min_bandwidth_fraction,
            high=self.max_bandwidth_fraction,
        )
E       AttributeError: module 'torch' has no attribute 'distribution'
```

keunwoochoi:
- I changed the default values and add relevant `raise` statements. 
- If I understood the expected args of LPF correctly, It'd also make sense to add this in the `__init__` of `BandPassFilter`. Would we? It'd make it easier to fix the value error.
```
        if sample_rate is not None:
            if max_center_frequency * max_bandwidth_fraction > 0.5 * sample_rate:
                raise ValueError(f"(some error message)")
```

iver56:
Ah, yes, julius doesn't allow the cutoff to exceed the nyquist frequency, right? I didn't think of that. Note that `sample_rate` can be passed either to `__init__` or to `forward` (or both, in which case the one given to forward will be used). If we want to be really thorough, the sanity check could be wrapped in a function that gets called from both `__init__` and `randomize_parameters`.

keunwoochoi:
Yup. For me it seems like a design choice. IMO the current version isn't too bad as i) we make it explicit and ii) in another error case with with a lower `sample_rate` passed in `forward()`, it'll still shows the error message from `julius`. That said, if you'd like the same sanity check in `forward()` I can add it.

iver56:
Nah, I guess we can get by without it for now :) Done is better than perfect

iver56:
Again, thanks for the contribution. I ran some audio through this and it sounds fine with the params you chose.

keunwoochoi:
Awesome!

PR #105, opened 2021-10-11
Make a shared requirements.txt for tests & build

KentoNishi:
Currently, updating requirements needs to be done in two files separately. Why not just make a shared requirements.txt file for the package dependencies and install them both in CI tests? I moved the `setup.py` requirements list into `requirements.txt` which is also referenced when initializing the environment in actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Install python dependencies
       run: |
         python -m pip install --upgrade --user pip --quiet
-        python -m pip install -r .github/workflows/test_requirements.txt
+        python -m pip install -r .github/workflows/test_requirements.txt -r requirements.txt
         python --version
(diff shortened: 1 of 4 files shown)


iver56:
I appreciate the idea of specifying the dependencies in a DRYer way, but I am hesitant to accept these changes. Here are my thoughts:

1) I want to keep using conda for the dev environment, because I've never been able to install a GPU-enabled pytorch without it. I feel like conda works well, so I haven't had the motivation to "downgrade" to using only pip
2) The package dependencies specified in setup.py must allow a wide range of dependency versions while the CI and dev environment should use pinned packages

If we should make the dependencies list DRYer, I feel like the CI should use conda and environment.yml (instead of pip and a test_requirements.txt) to set up its environment. Then we'll have two lists instead of three. But let's not boil it down to one.

PR #113, opened 2022-01-02
Implements SomeOf and OneOf

akashrajkn:
- Created a base class called `BaseCompose` similar to audiomentations
- `SomeOf` is a subclass of `BaseCompose` which randomizes the transforms applied on each call (and checks if number of transfroms to be applied is less than total number of transforms available)
- `OneOf` is a subclass of `SomeOf` which only applies one transform at a time

### DIFF
--- a/torch_audiomentations/core/composition.py
+++ b/torch_audiomentations/core/composition.py
@@ -92,7 +92,7 @@ class SomeOf(BaseCompose):
                  `SomeOf((1, 3), [transform1, transform2, transform3])`
 
         Example: Pick 2 to all of the transforms
-                 `SomeOf((1, None), [transform1, transform2, transform3])`
+                 `SomeOf((2, None), [transform1, transform2, transform3])`
     """
 
     def __init__(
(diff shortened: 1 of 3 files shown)


iver56:
Thanks for the pull request :) I just got back from vacation. I'll have a look at this within a couple of days

torch_audiomentations/core/composition.py
```diff
@@ -71,3 +60,94 @@ def supported_modes(self) -> set:
...
+      - Pick exactly n transforms
+        Example: pick exactly 2 of the transforms
+                 `SomeOf(2, [transform1, transform2, transform3])`
+
+      - Pick between a minimum and maximum number of transforms
+        Example: pick 1 to 3 of the transforms        
+                 `SomeOf((1, 3), [transform1, transform2, transform3])`
+
+        Example: Pick 2 to all of the transforms
+                 `SomeOf((1, None), [transform1, transform2, transform3])`
```
iver56, line 95:
```suggestion
                 `SomeOf((2, None), [transform1, transform2, transform3])`
```

so it matches the explanation in the line above

iver56 (changes requested):
The code looks good to me :) Good work. Just requesting one tiny change here

Issue #94, opened 2021-07-26
bug in adding background noise when noise is short, sr of input is 22050, and sr of noise is 16000


keunwoochoi:
In such a scenario, I'm facing this error.
```
    
        return selected_samples + background_rms.unsqueeze(-1) * background.view(
>           batch_size, 1, num_samples
        ).expand(-1, num_channels, -1)
E       RuntimeError: shape '[1, 1, 121809]' is invalid for input of size 121810


```

I suspect this is coming from an off-by-1 error introduced in `audio.get_num_samples()`, i.e.,

```
return round(num_samples * self.sample_rate / sample_rate)
``` 
, probably this rounding function. but i'm not very sure.

I confirmed that if I resample the noise file fro 16000 to 22050, the error disappears.

iver56:
Thanks for filing the bug! I thought we had fixed this already, but apparently it's still a problem :)

I think the best way to move forward is to first reproduce this bug in a test and then fix the bug.

When https://github.com/adefossez/julius/pull/14 (for avoiding off-by-one errors) is merged and released we might switch from librosa to julius, which is fast and CUDA-compatible.

PR #117, opened 2022-01-31
Explicit round to solve issue #94

MorenoLaQuatra:
Rounded value using math.ceil function to solve error in size for background noise.


iver56:
Hi :) Thanks for the PR.

Could you please add a unit test that covers this case (that failed before, but passes after your patch)? That will help against regressions

iver56:
This PR, when merged, closes #94

iver56:
This fix seems solid. I'll take care of the test.

PR #125, opened 2022-03-23
feat: add support for batches in Audio.rms_normalize

hbredin:
Audio.rms_normalize currently only works for (channel, time)-shaped samples. 

This PR adds support for (batch_size, channel, time)-shaped samples as well, which is the usual shape of `selected_samples` in `randomize_parameters` and `apply_transform`.


iver56:
Thanks for this :) I would love a unit test. If you don't have one, I can probably make one in the coming days

PR #126, opened 2022-03-25
Add ObjectDict class

iver56:


torch_audiomentations/utils/object_dict.py
```diff
@@ -0,0 +1,39 @@
+# Inspired by tornado
+# https://www.tornadoweb.org/en/stable/_modules/tornado/util.html#ObjectDict
+
+try:
+    import typing
+    from typing import cast
```
hbredin, line 6: Can you explain why this is imported? 
Is this a way to detect whether one can use `typing.Dict[...]` syntax?

iver56: Hmm, I'm not sure. I think it is a copy of tornado code from a time when it supported Python 3.4. This try except is probably not needed anymore, but it also doesn't hurt too much. Should I remove it?

hbredin: I would remove it, yes.
Note that I cherry-picked this PR in #123.

iver56: Done

hbredin (approved):

iver56:
Should I merge it, or will you cherry-pick it into your branch?

hbredin:
I cherry-picked it.

Issue #97, opened 2021-08-05
Implement a transform that crops audio to a given length


iver56:
The offset can be randomized, as long as the output has the specified length

Like https://github.com/Spijkervet/torchaudio-augmentations/blob/d044f9d020e12032ab9280acf5f34a337e72d212/torchaudio_augmentations/augmentations/random_resized_crop.py#L5

The idea is that one can have a chain of transforms, and some of them change the input length, but the final length should be fixed. That is where this transform is handy.

swagshaw:
@iver56  Hello, is this issue still opening? I would like to take it.

iver56:
Hi :) Yes, go ahead

shahules786:
@iver56 Do you want to close this issue?

PR #127, opened 2022-03-29
Dev#97

shahules786:
PR to implement crop transform https://github.com/asteroid-team/torch-audiomentations/issues/97

### DIFF
--- a/torch_audiomentations/augmentations/random_crop.py
+++ b/torch_audiomentations/augmentations/random_crop.py
@@ -7,14 +7,30 @@ from ..core.transforms_interface import MultichannelAudioNotSupportedException
 
 class RandomCrop(torch.nn.Module):
 
-    """Crop the audio to predefined length in seconds."""
+    """Crop the audio to predefined length in max_length."""
 
     supports_multichannel = True
 
-    def __init__(self, seconds: float, sampling_rate: int):
+    def __init__(self, 
+        max_length: float, 
+        sampling_rate: int, 
+        max_length_unit:str = "seconds"
+    ):
+        """
+        :param max_length: length to which samples are to be cropped.
+        :sampling_rate: sampling rate of input samples.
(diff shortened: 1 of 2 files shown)


iver56:
I believe a common use case for this transform will be to adapt the audio to the constant length requirement of a ML model. Such length requirements can be oddly specific, for example 461906 samples. I would like to have an option to specify the desired output length in terms of samples. Number of seconds also makes sense in some cases :)

Maybe you could add support for specifying the length in samples, just like in `Shift`?

https://github.com/asteroid-team/torch-audiomentations/blob/master/torch_audiomentations/augmentations/shift.py#L69

shahules786:
Yes, I have added the argument @iver56 .

torch_audiomentations/augmentations/random_crop.py
```diff
@@ -7,14 +7,30 @@
...
+        """
         super(RandomCrop, self).__init__()
         self.sampling_rate = sampling_rate
-        self.num_samples = int(self.sampling_rate * seconds)
+        if max_length_unit == "seconds":
+            self.num_samples = int(self.sampling_rate * max_length)
+        elif max_length_unit == "samples":
+            self.num_samples = int(max_length)
+        else:
+            raise ValueError('max_length must be "samples" or "seconds"')
```
iver56, line 33:
```suggestion
            raise ValueError('max_length_unit must be "samples" or "seconds"')
```

shahules786: Thanks @iver56 , done.

iver56:
Thanks! I'm not sure if I can squeeze this into the upcoming release, as there will also be a lot going on with https://github.com/asteroid-team/torch-audiomentations/pull/123 that changes the API of things

But I can at least merge it now, and then maybe massage it (like add target support) and add it to the API later 👍

PR #129, opened 2022-04-04
feat: add "output_type" argument

hbredin:
... as discussed [here](https://github.com/asteroid-team/torch-audiomentations/pull/123#issuecomment-1085884342).


hbredin:
Note that `RandomCrop` has been added after the `ObjectDict` PR and should be updated...

iver56 (approved):
Thanks 💯

PR #130, opened 2022-04-05
feat: add Identity transform

hbredin:
This PR adds the `Identity` transform that basically does nothing (i.e. returns its input unchanged).

One benefit is to simplify codebase of other libraries  using `torch_audiomentations` (pyannote.audio, I am looking at you ;)):

Before:

```python
if use_augmentation:
    augmentation = MyAugmentation(...)

for samples in dataloader:

  if use_augmentation:
     augmented = augmentation(samples, ...)
  else:
     augmented = samples
```

After: 

```python
augmentation = MyAugmentation(...) if use_augmentation else Identity()

for samples in dataloader:
   augmented = augmentation(samples, ...)
```


iver56:
Sure, why not ^^
Thanks for the contribution

PR #131, opened 2022-04-05
feat: add support for external augmentation in from_dict

hbredin:
`torch_audiomentations.utils.config.from_dict` only allows to instantiate `torch_audiomentations` built-in augmentations.

This PR brings support for augmentations provided by other library (e.g. augmentation so  esoteric that they would never be merged in `torch_audiomentations` but that you still want to compose with built-in `torch_audiomentations` augmentations).

**Before:**

```python
compose = from_dict(
   {'transform': 'Compose',
    'params': {'transforms': [{'transform': 'Gain'},
                              {'transform': 'my_package.my_module.MyAugmentation'}]}})
```

would fail because it actually tries to load `torch_audiomentations.my_package.my_module.MyAugmentation`

**After:** it would actually work.

The alternative is to do something like this in `my_package`:

```python
import torch_audiomentations
from my_package.my_module import MyAugmentation
torch_audiomentations.MyAugmentation = MyAugmentation
```

@iver56 Feel free to just close this PR  if you think this should be handled by the said other library.

### DIFF
--- a/torch_audiomentations/utils/config.py
+++ b/torch_audiomentations/utils/config.py
@@ -1,5 +1,6 @@
 from pathlib import Path
-from typing import Any, Dict, Text, Union
+from typing import Any, Dict, Text, Optional, Union
+from importlib import import_module
 
 import torch_audiomentations
 from torch_audiomentations import Compose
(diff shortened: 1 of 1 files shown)


iver56:
LGTM 👍 

A related thing I thought about once was adding support for audiomentations transforms. For example, Mp3Compression can be useful, but we won't implement mp3 compression in pytorch like the usual torch-audiomentations transforms ^^

Issue #107, opened 2021-10-24
Implement SpliceOut


iver56:
https://arxiv.org/abs/2110.00046

shahules786:
Hey @iver56, I would like to look into this.

iver56:
Great! Go ahead 😄

iver56:
https://github.com/asteroid-team/torch-audiomentations/issues/140

PR #135, opened 2022-04-15
Spliceout transform

shahules786:
Implementation of spliceout transform, silence padding is added at the end. https://github.com/asteroid-team/torch-audiomentations/issues/107

### DIFF
--- a/tests/test_spliceout.py
+++ b/tests/test_spliceout.py
@@ -23,6 +23,21 @@ class TestSpliceout(unittest.TestCase):
 
         assert spliceout_samples.dtype == np.float32
 
+    def test_spliceout_odd_hann(self):
+
+        audio_samples = torch.rand(size=(8, 1, 32000), dtype=torch.float32)
+        augment = Compose(
+            [
+                SpliceOut(num_time_intervals=10, max_width=400, output_type="dict"),
+            ],
+            output_type="dict",
+        )
+        spliceout_samples = augment(
+            samples=audio_samples, sample_rate=16100
+        ).samples.numpy()
+
+        assert spliceout_samples.dtype == np.float32
+
     def test_spliceout_perbatch(self):
 
         audio_samples = torch.rand(size=(8, 1, 32000), dtype=torch.float32)
--- a/tests/test_spliceout.py
+++ b/tests/test_spliceout.py
@@ -13,7 +13,7 @@ class TestSplitout(unittest.TestCase):
         audio_samples = torch.rand(size=(8, 1, 32000), dtype=torch.float32)
         augment = Compose(
             [
-                SpliceOut(num_time_intervals=10, max_width=400, output_type="dict"),
+                SpliceOut(num_time_intervals=10, max_width=25, output_type="dict"),
             ],
             output_type="dict",
         )
@@ -30,7 +30,7 @@ class TestSplitout(unittest.TestCase):
             [
                 SpliceOut(
                     num_time_intervals=10,
-                    max_width=400,
+                    max_width=25,
                     mode="per_batch",
                     p=1.0,
                     output_type="dict",
@@ -51,7 +51,7 @@ class TestSplitout(unittest.TestCase):
         audio_samples = torch.rand(size=(8, 2, 32000), dtype=torch.float32)
         augment = Compose(
             [
-                SpliceOut(num_time_intervals=10, max_width=400, output_type="dict"),
+                SpliceOut(num_time_intervals=10, max_width=25, output_type="dict"),
             ],
             output_type="dict",
         )
@@ -74,7 +74,7 @@ class TestSplitout(unittest.TestCase):
         )
         augment = Compose(
             [
-                SpliceOut(num_time_intervals=10, max_width=400, output_type="dict"),
+                SpliceOut(num_time_intervals=10, max_width=25, output_type="dict"),
             ],
             output_type="dict",
         )
--- a/tests/test_spliceout.py
+++ b/tests/test_spliceout.py
@@ -13,7 +13,7 @@ class TestSplitout(unittest.TestCase):
         audio_samples = torch.rand(size=(8, 1, 32000), dtype=torch.float32)
         augment = Compose(
             [
-                SpliceOut(num_time_intervals=10, max_width=100, output_type="dict"),
+                SpliceOut(num_time_intervals=10, max_width=400, output_type="dict"),
             ],
             output_type="dict",
         )
@@ -30,7 +30,7 @@ class TestSplitout(unittest.TestCase):
             [
                 SpliceOut(
                     num_time_intervals=10,
-                    max_width=100,
+                    max_width=400,
                     mode="per_batch",
                     p=1.0,
                     output_type="dict",
@@ -51,7 +51,7 @@ class TestSplitout(unittest.TestCase):
         audio_samples = torch.rand(size=(8, 2, 32000), dtype=torch.float32)
         augment = Compose(
             [
-                SpliceOut(num_time_intervals=10, max_width=100, output_type="dict"),
+                SpliceOut(num_time_intervals=10, max_width=400, output_type="dict"),
             ],
             output_type="dict",
         )
@@ -74,7 +74,7 @@ class TestSplitout(unittest.TestCase):
         )
         augment = Compose(
             [
-                SpliceOut(num_time_intervals=10, max_width=100, output_type="dict"),
+                SpliceOut(num_time_intervals=10, max_width=400, output_type="dict"),
             ],
             output_type="dict",
         )
--- a/tests/test_spliceout.py
+++ b/tests/test_spliceout.py
@@ -1,0 +1,70 @@
+import unittest
+import numpy as np
+import torch
+import pytest
+
+from torch_audiomentations.augmentations.spliceout import SpliceOut
+from torch_audiomentations import Compose
+
+
+class TestSplitout(unittest.TestCase):
+
+    def test_spliceout(self):
+
+        audio_samples = torch.rand(size=(8, 1, 32000), dtype=torch.float32)
+        augment = Compose([
+                    SpliceOut(num_time_intervals=10,max_width=100,output_type='dict'),
+                ],output_type='dict')
+        spliceout_samples = augment(samples=audio_samples, sample_rate=16000).samples.numpy()
+
+        assert spliceout_samples.dtype == np.float32
+
+    def test_spliceout_perbatch(self):
+
+        audio_samples = torch.rand(size=(8, 1, 32000), dtype=torch.float32)
+        augment = Compose([
+                    SpliceOut(num_time_intervals=10,max_width=100, mode = "per_batch", p=1.0, output_type='dict'),
+                ],output_type='dict')
+        spliceout_samples = augment(samples=audio_samples, sample_rate=16000).samples.numpy()
+        
+        assert spliceout_samples.dtype == np.float32
+        self.assertLess(spliceout_samples.sum(), audio_samples.numpy().sum())
+        self.assertEqual(spliceout_samples.shape, audio_samples.shape)
+
+    def test_spliceout_multichannel(self):
+
+        audio_samples = torch.rand(size=(8, 2, 32000), dtype=torch.float32)
+        augment = Compose([
+                    SpliceOut(num_time_intervals=10,max_width=100, output_type='dict'),
+                ],output_type='dict')
+        spliceout_samples = augment(samples=audio_samples, sample_rate=16000).samples.numpy()
+        
+        assert spliceout_samples.dtype == np.float32
+        self.assertLess(spliceout_samples.sum(), audio_samples.numpy().sum())
+        self.assertEqual(spliceout_samples.shape, audio_samples.shape)
+
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="Requires CUDA")
+    def test_spliceout_cuda(self):
+        
+        audio_samples = torch.rand(size=(8, 1, 32000), dtype=torch.float32, device= torch.device('cuda')) - 0.5
+        augment = Compose([
+                    SpliceOut(num_time_intervals=10,max_width=100,output_type='dict'),
+                ],output_type='dict')
+        spliceout_samples = augment(samples=audio_samples, sample_rate=16000).samples.cpu().numpy()
+
+        assert spliceout_samples.dtype == np.float32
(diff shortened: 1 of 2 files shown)


iver56:
Thanks for the PR, and thanks for the patience 😃

I had a look at this now, and I have a concern about this: You are using stft and istft. I can see where that is coming from, as the spliceout paper demonstrates the technique with an illustration with spectrograms. However, as far as I know there is no good reason for doing the splices in the spectrogram domain - it can be done directly on the waveform with a similar result, and that would be much faster, use less memory and need less code.

Would you consider rewriting it to work directly on the waveform? Or, alternatively, do you have a good reason for going via the spectrogram?

iver56 (changes requested):
See comment above

shahules786:
> See comment above

shahules786:
Hi @iver56, you're right. We can do it directly on the time domain. I have updated the code.

torch_audiomentations/augmentations/spliceout.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+
+    supported_modes = {"per_batch", "per_example"}
+
+    def __init__(
+        self,
+        num_time_intervals,
+        max_width,
+        mode: str = "per_example",
+        n_fft=400,
```
iver56, line 26: This param is no longer used. It should be removed

torch_audiomentations/augmentations/spliceout.py
```diff
@@ -0,0 +1,98 @@
...
+    spliceout augmentation proposed in https://arxiv.org/pdf/2110.00046.pdf
+    silence padding is added at the end to retain the audio length.
+    """
+
+    supported_modes = {"per_batch", "per_example"}
+
+    def __init__(
+        self,
+        num_time_intervals,
+        max_width,
```
iver56, line 24: Most transforms in audiomentations and torch-audiomentations provide sane defaults as a good starting point, and I'd love to see that here too.

Based on what I read in the paper I would suggest num_time_intervals = 8 and max_width = 400 ms as default. The problem with max_width is that it is currently defined as the number of samples, so it will change based on sample rate. Maybe it makes more sense to specify it in seconds by default? Other transforms, like `Shift`, have an extra parameter that defines the unit (samples, seconds or fraction). Maybe that's a useful pattern to adopt here as well?

https://github.com/asteroid-team/torch-audiomentations/blob/master/torch_audiomentations/augmentations/shift.py#L81

torch_audiomentations/augmentations/spliceout.py
```diff
@@ -0,0 +1,98 @@
...
+                    samples[i].shape[-1] - random_lengths[j], size=(1,)
+                )
+                mask[start : start + random_lengths[j]] = False
+
+            spliceout_sample = samples[i][:,mask]
+            padding = torch.zeros(
+                (samples[i].shape[0], samples[i].shape[-1] - spliceout_sample.shape[-1]),
+                dtype=torch.float32,
+            )
+            spliceout_sample = torch.cat((spliceout_sample, padding), dim=-1).unsqueeze(0)
```
iver56, line 90: One concern I have regarding this implementation is clicks/pops that will occur as a result of the slicing/splicing/joining the audio waveform. It can be alleviated by having a very short (like maybe 10 ms) crossfade at the joints, maybe with a hann window or with sqrt?

I've tried to illustrate the problem here, with a low-frequency sine wave:
bilde

As you can see, the problem is that the joining the parts like this can create a rather loud impulse/click. Here it is as an audio file, so you can also hear it:

[low_freq_sine.zip](https://github.com/asteroid-team/torch-audiomentations/files/8553714/low_freq_sine.zip)

iver56 (changes requested):

torch_audiomentations/augmentations/spliceout.py
iver56: I tried to illustrate the crossfading idea:

bilde

torch_audiomentations/augmentations/spliceout.py
```diff
@@ -0,0 +1,124 @@
...
+    silence padding is added at the end to retain the audio length.
+    """
+
+    supported_modes = {"per_batch", "per_example"}
+    requires_sample_rate = True
+
+    def __init__(
+        self,
+        num_time_intervals=8,
+        max_width=400,
```
iver56, line 26: I would still prefer to have this specified in seconds instead of samples, because this default is going to behave different based on the sample rate of the input. For example, 400 samples is 25 milliseconds with 16 kHz sample rate, but ~8 milliseconds with 48 kHz

iver56 (changes requested):

iver56:
I ran `pytest` on my end and observed a test failure:

```
_____________________________________________________________ TestSplitout.test_spliceout_cuda _____________________________________________________________

self = <tests.test_spliceout.TestSplitout testMethod=test_spliceout_cuda>

    @pytest.mark.skipif(not torch.cuda.is_available(), reason="Requires CUDA")
    def test_spliceout_cuda(self):

        audio_samples = (
            torch.rand(
                size=(8, 1, 32000), dtype=torch.float32, device=torch.device("cuda")
            )
            - 0.5
        )
        augment = Compose(
            [
                SpliceOut(num_time_intervals=10, max_width=25, output_type="dict"),
            ],
            output_type="dict",
        )
        spliceout_samples = (
>           augment(samples=audio_samples, sample_rate=16000).samples.cpu().numpy()
        )

tests\test_spliceout.py:82:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
..\..\anaconda3\envs\torch-audiomentations-gpu\lib\site-packages\torch\nn\modules\module.py:1110: in _call_impl
    return forward_call(*input, **kwargs)
torch_audiomentations\core\composition.py:120: in forward
    inputs = self.transforms[i](**inputs)
..\..\anaconda3\envs\torch-audiomentations-gpu\lib\site-packages\torch\nn\modules\module.py:1110: in _call_impl
    return forward_call(*input, **kwargs)
torch_audiomentations\core\transforms_interface.py:341: in forward
    perturbed: ObjectDict = self.apply_transform(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = SpliceOut()
samples = tensor([[[-0.2683,  0.4740,  0.1533,  ..., -0.2279, -0.2688,  0.0713]],

        [[ 0.4288,  0.3069,  0.1189,  ..., -0...,  0.4776,  0.0145]],

        [[ 0.4209,  0.2337,  0.0237,  ...,  0.2466,  0.0779,  0.1981]]],
       device='cuda:0')
sample_rate = 16000, targets = None, target_rate = None

    def apply_transform(
        self,
        samples: Tensor = None,
        sample_rate: Optional[int] = None,
        targets: Optional[Tensor] = None,
        target_rate: Optional[int] = None,
    ) -> ObjectDict:

        spliceout_samples = []
        hann_window_len = int(sample_rate * 0.02)
        hann_window = torch.hann_window(hann_window_len)
        hann_window_left, hann_window_right = (
            hann_window[: hann_window_len // 2],
            hann_window[hann_window_len // 2 :],
        )

        for i in range(samples.shape[0]):

            random_lengths = self.transform_parameters["splice_lengths"][i]
            mask = torch.ones(samples[i].shape[-1], dtype=bool)
            all_starts = []

            for j in range(self.num_time_intervals):
                start = torch.randint(
                    int(sample_rate * 0.01),
                    samples[i].shape[-1] - random_lengths[j],
                    size=(1,),
                )
                mask[start : start + random_lengths[j]] = False
                all_starts.append(start)

            spliceout_sample = samples[i][:, mask]
            padding = torch.zeros(
                (samples[i].shape[0], samples[i].shape[-1] - spliceout_sample.shape[-1]),
                dtype=torch.float32,
            )
>           spliceout_sample = torch.cat((spliceout_sample, padding), dim=-1)
E           RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument tensors in method wrapper___cat)
```

tests/test_spliceout.py
```diff
@@ -0,0 +1,86 @@
+import unittest
+import numpy as np
+import torch
+import pytest
+
+from torch_audiomentations.augmentations.spliceout import SpliceOut
+from torch_audiomentations import Compose
+
+
+class TestSplitout(unittest.TestCase):
```
iver56, line 10:
```suggestion
class TestSpliceout(unittest.TestCase):
```

iver56:
Do you need/want help with fixing the CUDA test case? Or do you have a CUDA-capable setup now?

iver56:
I tried to include this transform in the demo script, but it failed with an error message like this:

```
Traceback (most recent call last):
  File "C:\Users\iver5\Anaconda3\envs\torch-audiomentations-gpu\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\iver5\Anaconda3\envs\torch-audiomentations-gpu\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\iver5\Code\torch-audiomentations\scripts\demo.py", line 189, in <module>
    augmented_samples = augmenter(
  File "C:\Users\iver5\Anaconda3\envs\torch-audiomentations-gpu\lib\site-packages\torch\nn\modules\module.py", line 1110, in _call_impl
    return forward_call(*input, **kwargs)
  File "C:\Users\iver5\Code\torch-audiomentations\torch_audiomentations\core\transforms_interface.py", line 454, in forward
    perturbed: ObjectDict = self.apply_transform(
  File "C:\Users\iver5\Code\torch-audiomentations\torch_audiomentations\augmentations\spliceout.py", line 113, in apply_transform
    spliceout_sample[
RuntimeError: The expanded size of the tensor (440) must match the existing size (441) at non-singleton dimension 1.  Target sizes: [4, 440].  Tensor sizes: [4, 441]
```

The branch is here:
https://github.com/asteroid-team/torch-audiomentations/tree/ij/spliceout-demo

And the demo script can be run like this:

`python -m scripts.demo`

Could you have a look at this?

torch_audiomentations/augmentations/spliceout.py
```diff
@@ -0,0 +1,124 @@
...
+    silence padding is added at the end to retain the audio length.
+    """
+
+    supported_modes = {"per_batch", "per_example"}
+    requires_sample_rate = True
+
+    def __init__(
+        self,
+        num_time_intervals=8,
+        max_width=25,
```
iver56, line 26: Ideally, this default should be 400 ms, not 25 ms. The motivation here is that the authors of the paper used T=40 successfully, which in their case means 40 * 10 ms (because their spectrogram was made with a hop size of 10 ms)

https://arxiv.org/pdf/2110.00046.pdf

iver56 (changes requested):

torch_audiomentations/augmentations/spliceout.py
shahules786: Done.

tests/test_spliceout.py
shahules786: Done

shahules786:
> Do you need/want help with fixing the CUDA test case? Or do you have a CUDA-capable setup now?
I don't have it yet. I hope I have resolved this with the latest commit but if the error prevails feel free to do a fix.

shahules786:
> I tried to include this transform in the demo script, but it failed with an error message like this:
> 
> ```
> Traceback (most recent call last):
>   File "C:\Users\iver5\Anaconda3\envs\torch-audiomentations-gpu\lib\runpy.py", line 194, in _run_module_as_main
>     return _run_code(code, main_globals, None,
>   File "C:\Users\iver5\Anaconda3\envs\torch-audiomentations-gpu\lib\runpy.py", line 87, in _run_code
>     exec(code, run_globals)
>   File "C:\Users\iver5\Code\torch-audiomentations\scripts\demo.py", line 189, in <module>
>     augmented_samples = augmenter(
>   File "C:\Users\iver5\Anaconda3\envs\torch-audiomentations-gpu\lib\site-packages\torch\nn\modules\module.py", line 1110, in _call_impl
>     return forward_call(*input, **kwargs)
>   File "C:\Users\iver5\Code\torch-audiomentations\torch_audiomentations\core\transforms_interface.py", line 454, in forward
>     perturbed: ObjectDict = self.apply_transform(
>   File "C:\Users\iver5\Code\torch-audiomentations\torch_audiomentations\augmentations\spliceout.py", line 113, in apply_transform
>     spliceout_sample[
> RuntimeError: The expanded size of the tensor (440) must match the existing size (441) at non-singleton dimension 1.  Target sizes: [4, 440].  Tensor sizes: [4, 441]
> ```
> 
> The branch is here: https://github.com/asteroid-team/torch-audiomentations/tree/ij/spliceout-demo
> 
> And the demo script can be run like this:
> 
> `python -m scripts.demo`
> 
> Could you have a look at this?

I have fixed this.

iver56:
Thanks for the patience :) I tried running the demo script again, and it runs without exceptions now.

`python -m scripts.demo`

However, I listened to the result and there are still some impulse artifacts. It sounds like little clicks. Maybe there's still a bug in the crossfading code somehow?

It looks like this in the waveform:

bilde

And one can see these vertical lines that span the whole frequency spectrum when visualizing it with a spectrogram:

bilde

Here is a copy of the example I evaluated here:

[SpliceOut_per_example_000_perfect-alley1.zip](https://github.com/asteroid-team/torch-audiomentations/files/8618656/SpliceOut_per_example_000_perfect-alley1.zip)

iver56:
If you merge this, you can also run the demo script in your branch:

https://github.com/shahules786/torch-audiomentations/pull/1

iver56:
You might have to rebase your branch against master as well, as I have added some fixes to the demo script to make it robust to the output type changes that are still in limbo in master

shahules786:
Spectrogram from latest commit 
Screenshot 2022-05-04 at 11 04 14 PM

iver56 (approved):
Thanks for the contribution 👍

PR #138, opened 2022-04-21
Fix a bug where BandPassFilter didn't work on GPU

### LINKED ISSUE #137: [Bug] RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu
Hi, 
It happens in band_pass_filter.py 
_bandwidth_dist_ is allocated on the CPU (ln 101-103) while _center_dist_ (ln96-98) is allocated on the GPU.
I think that it can be solved by using the device info of _selected_samples_ for _bandwidth_dist_ .

iver56:
Close #137

### DIFF
--- a/tests/test_band_pass_filter.py
+++ b/tests/test_band_pass_filter.py
@@ -1,4 +1,5 @@
 import numpy as np
+import pytest
 import torch
 
 from torch_audiomentations import BandPassFilter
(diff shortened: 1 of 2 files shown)


ahikaml:
Just one comment - it should be selected_samples.device instead of samples.device

iver56:
I see where that comment is coming from - torch-audiomentations is in flux, and that variable has been renamed lately:
https://github.com/asteroid-team/torch-audiomentations/commit/4ec0aaf00db5bf70facf23dd7cc2b61c8585fb5a#diff-1cf4d3614a67d1ef0ec702b0927ec9934de0c88abb1c8e90bf07321d631037eeR83

This change is not released yet

ahikaml:
Got it. Thanks

PR #139, opened 2022-05-01
Vocal tract length perturbation

### LINKED ISSUE #115: Implement Vocal Tract Length Perturbation
It was introduced here: https://arxiv.org/pdf/2011.12536.pdf
VTLP is used in speech recognition (for example: https://arxiv.org/pdf/2201.05562.pdf)

Reference code for vtlp is available here: https://github.com/makcedward/nlpaug/blob/master/nlpaug/model/audio/vtlp.py

akashrajkn:
Implemented VLTP as introduced in http://www.cs.toronto.edu/~hinton/absps/perturb.pdf. Adopted from the numpy code here: https://github.com/makcedward/nlpaug/blob/master/nlpaug/model/audio/vtlp.py

Additional notes:
- Only supports single channel for now
- At the moment, the entire batch will use one (randomized) warp factor
- Added a sanity test case

Fixes #115

### DIFF
--- a/tests/test_vtlp.py
+++ b/tests/test_vtlp.py
@@ -1,7 +1,7 @@
 import numpy as np
 import torch
 
-from torch_audiomentations import VTLP
+from torch_audiomentations import VocalTractLengthPerturbation
 
 
 class TestVTLP:
@@ -9,7 +9,7 @@ class TestVTLP:
         samples = torch.randn(3, 1, 32000)
         sample_rate = 16000
 
-        augment = VTLP(p=1.0, sample_rate=sample_rate, output_type="dict")
+        augment = VocalTractLengthPerturbation(p=1.0, sample_rate=sample_rate, output_type="dict")
         processed_samples = augment(
             samples=samples, sample_rate=sample_rate
         ).samples.numpy()
--- a/tests/test_vtlp.py
+++ b/tests/test_vtlp.py
@@ -1,0 +1,18 @@
+import numpy as np
+import torch
+
+from torch_audiomentations import VTLP
+
+
+class TestVTLP:
+    def test_vtlp(self):
+        samples = torch.randn(3, 1, 32000)
+        sample_rate = 16000
+
+        augment = VTLP(p=1.0, sample_rate=sample_rate, output_type="dict")
+        processed_samples = augment(
+            samples=samples, sample_rate=sample_rate
+        ).samples.numpy()
+
+        assert processed_samples.shape == samples.shape
+        assert processed_samples.dtype == np.float32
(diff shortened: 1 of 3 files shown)


iver56:
Thanks for the pull request :) I'll find time to review this

iver56 (changes requested):
Interesting! I will have another look at this soon, but I've left some initial comments.

I also want to add a mono voice recording in the demo, so I can try it on that.

I understand that this transform does not support stereo/multichannel audio yet. Is that right? Would it be hard to add support for that? The other transforms in torch-audiomentation support multichannel audio

iver56:
I added VTLP to the demo script and added a speech example there:
https://github.com/akashrajkn/torch-audiomentations/pull/1

I listened to the outputs, and it very much resembles a band stop filter. Is that what you intended? I've attached the sounds in this zip:

[vtlp_demo_output.zip](https://github.com/asteroid-team/torch-audiomentations/files/8638202/vtlp_demo_output.zip)

And I have another question: is this a different technique? https://www.isca-speech.org/archive/pdfs/interspeech_2019/kim19_interspeech.pdf

iver56:
vtlp_spectrograms

iver56:
I'm trying to learn about what Vocal Tract Length Perturbation is. Reading the paper, I get the idea that it's about frequency warping

> For VTLP, we generate a random warp factor α for each utterance, and warp the freqency axis, such that a frequency f is mapped to a new frequency f ′

But I don't see frequencies getting re-mapped in the spectrogram gif I posted above

I also found this video online:

https://www.youtube.com/watch?v=vCDnfUM6gn8

Could you try to enlighten me? Do you have some reference examples of what VTLP should sound like? Am I missing something, or is there a bug in your implementation?

akashrajkn:
> I understand that this transform does not support stereo/multichannel audio yet. Is that right?

- [ ] That is right. I will modify it to support stereo audio

> Is that the case here, or is it enough to just use the sample_rate you get from forward calls?

- [ ] You are right, it is not required in the init function. I'll change it accordingly

> And I have another question: is this a different technique? https://www.isca-speech.org/archive/pdfs/interspeech_2019/kim19_interspeech.pdf

I wasn't aware of this paper, I will read it and get back to you

> Do you have some reference examples of what VTLP should sound like?

- [ ] I postponed adding tests in this PR. I will add some reference examples

> Am I missing something, or is there a bug in your implementation?

I think maybe there is an issue with my implementation. I will check it out and comment here (probably on the weekend)

iver56:
Closing for inactivity. Feel free to suggest a reopen later if the work gets picked up again

PR #142, opened 2022-05-12
Noise gate augmentation

shahules786:
Implemented spectral gating based on https://timsainburg.com/noise-reduction-python.html
Can you check this and let me know what you think? @iver56 
I will implement the tests after the review if that's fine.

### DIFF
--- a/torch_audiomentations/augmentations/noise_gates.py
+++ b/torch_audiomentations/augmentations/noise_gates.py
@@ -1,0 +1,48 @@
+import torch
+from torch import Tensor
+from typing import Optional
+
+from ..core.transforms_interface import BaseWaveformTransform
+from torch_audiomentations.utils.object_dict import ObjectDict
+
+
+
+class SpectralGating(BaseWaveformTransform):
+
+    def __init__(
+        self,
+        std_away = 1.0,
+        n_grad_freq=2,
+        n_grad_time=4,
+        decrease_prop=1,
+        n_fft=2048,
+        win_length=2048,
+        hop_length=512,
+        mode: str = "per_example",
+        p: float = 0.5,
+        p_mode: Optional[str] = None,
+        sample_rate: Optional[int] = None,
+        target_rate: Optional[int] = None,
+        output_type: Optional[str] = None,
+    ):
+        super(SpectralGating, self).__init__()
+
+
+    def randomize_parameters(
+        self,
+        samples: Tensor = None,
+        sample_rate: Optional[int] = None,
+        targets: Optional[Tensor] = None,
+        target_rate: Optional[int] = None):
+
+        pass
+
+    def apply_transform(
+        self,
+        samples: Tensor = None,
+        sample_rate: Optional[int] = None,
+        targets: Optional[Tensor] = None,
+        target_rate: Optional[int] = None) -> ObjectDict:
+
+        pass
+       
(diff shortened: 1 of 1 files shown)


torch_audiomentations/augmentations/noise_gates.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def apply_transform(
+        self,
+        samples: Tensor = None,
+        sample_rate: Optional[int] = None,
+        targets: Optional[Tensor] = None,
+        target_rate: Optional[int] = None) -> ObjectDict:
+
+        ### threshold 
+        if self.mode == "per_batch":
```
iver56, line 71: What was the motivation for having separate code for per_batch, per_example and per_channel? Normally the BaseWaveformTransform will take care of reshaping the audio tensor to a "per example" kind so the transform doesn't have to deal with each mode separately.

iver56 (changes requested):
Thanks for the initiative! It would be nice if you could add it to the demo script as well, so we can listen to what it sounds like :)

torch_audiomentations/augmentations/noise_gates.py
shahules786: To calculate the noise threshold differently for each mode. For example, if mode = "per_example", the noise threshold is set differently for each example and processed accordingly. You can see the code for each one is a bit different from others.

iver56: I see. But I get the feeling that the mode is taken into acount "twice", as in it gets handled first by the base class (reshaping) and then also in this transform. I haven't studied the code in detail though.

The base class will reshape the audio tensor. For example:
* per_example: `(num_examples, num_channels, samples)`
* per_batch: `(1, num_examples * num_channels, samples)`
* per_channel: `num_examples * num_channels, 1, samples)`

So the idea is that the transform can always assume a "per_example" way of transforming the audio for simplicity

shahules786:
> Thanks for the initiative! It would be nice if you could add it to the demo script as well, so we can listen to what it sounds like :)

Done.

torch_audiomentations/augmentations/noise_gates.py
shahules786: @iver56 got it. I have made the changes to remove mode-based processing. I have added the function to the demo as well. Let me know what you think.

iver56:
I listened to an output from the demo script now, and I can hear a buzzy sound that was somehow added to my original sound. I did not expect this. Maybe something is wrong with the implementation? I can also see it in the spectrogram:
spectralgating

I expected something more like this:

spectralgating-expected

(I created this "expected" example with Audacity's built-in noise reduction plugin)

Here are the 3 wavs that I compared in those two spectrogram gifs:
[spectral_gating_sounds.zip](https://github.com/asteroid-team/torch-audiomentations/files/8717206/spectral_gating_sounds.zip)

shahules786:
> I listened to an output from the demo script now, and I can hear a buzzy sound that was somehow added to my original sound. I did not expect this. Maybe something is wrong with the implementation? I can also see it in the spectrogram: spectralgating [ ](https://user-images.githubusercontent.com/1470603/169065982-9f79a357-6e02-4443-bfa3-5efe617f8727.gif) spectralgating [ ](https://user-images.githubusercontent.com/1470603/169065982-9f79a357-6e02-4443-bfa3-5efe617f8727.gif)
> 
> I expected something more like this:
> 
> spectralgating-expected [ ](https://user-images.githubusercontent.com/1470603/169067794-ac466d50-8cc2-4db3-a27a-f51393a66412.gif) spectralgating-expected [ ](https://user-images.githubusercontent.com/1470603/169067794-ac466d50-8cc2-4db3-a27a-f51393a66412.gif)
> 
> (I created this "expected" example with Audacity's built-in noise reduction plugin)
> 
> Here are the 3 wavs that I compared in those two spectrogram gifs: [spectral_gating_sounds.zip](https://github.com/asteroid-team/torch-audiomentations/files/8717206/spectral_gating_sounds.zip)

I too observed that. Thanks for the reference spectrogram.

Issue #140, opened 2022-05-06
Improve Spliceout


iver56:
Here are some ideas Spliceout improvements:
- [ ] Add a param for crossfade window size? 10 ms may actually be a bit on the large side for a default
- [ ] Maybe try sqrt for crossfades instead of hann to avoid the perceived dip in energy in the crossfades
- [ ] Improve the docstring (ELI5)
- [ ] Rename max_width? And specify it in seconds instead of milliseconds.
- [ ] Run `black` on the code

PR #143, opened 2022-05-27
add SpliceOut to torch_audiomentations init

### LINKED ISSUE #144: SpliceOut can not be used with `from_yaml`
SpliceOut is not in the main `__init__.py` file and can therefore not be used when loading a configuration using e.g. `from_yaml`

HLasse:
Since SpliceOut is not in the main `__init__.py` fiie, it can not be used when loading a configuration using e.g. `from_yaml`. Fixes #144


iver56:
Thanks for the PR, and your interest in SpliceOut! SpliceOut isn't "officially released" yet, so that's why I haven't added it to the API yet. There are some things I still want to change about it

https://github.com/asteroid-team/torch-audiomentations/issues/140

Issue #172, opened 2024-01-10
Remove soundfile backend setting


MorenoLaQuatra:
The `soundfile` backend seems to be the default. According to some new UserWarning:

```
MY_INSTALL_FOLDER/torch_audiomentations/utils/io.py:27: UserWarning: torchaudio._backend.set_audio_backend has been deprecated. With dispatcher enabled, this function is no-op. You can remove the function call
```
it should be possible to remove this line https://github.com/asteroid-team/torch-audiomentations/blob/0dea5dbe552e974e9f4e5034924a253cae72351e/torch_audiomentations/utils/io.py#L27

PR #173, opened 2024-01-10
Removing set_backend to avoid torchaudio UserWarning

MorenoLaQuatra:
Commented out the `set_backend` to avoid `torchaudio` UserWarning.

Ref. #172


iver56:
This change is included in release 0.11.1

PR #182, opened 2025-01-14
chore: remove librosa dependency in favor of torchaudio

hbredin:
librosa depends on numba that is a dependency hell.
Since torchaudio is already a dependency, I thought it would be nice to use it fully :)


iver56:
I wholeheartedly agree. As a nice bonus, this will make it faster too. Thanks for the contribution!

hbredin:
🎉

PR #183, opened 2025-06-27
Add RMSNormalization transform for target-level RMS normalisation (closes #176)

### LINKED ISSUE #176: Feature Request: Support for Target Level Normalization
## Feature Description
I propose the addition of an target level normalization augmentation to the `torch-audiomentations` library. This feature would analyze an audio signal's average power and adjust its gain to normalize the signal to a target RMS level.

## Current Limitations
While the library currently supports peak normalization and gain adjustments, these methods do not directly allow setting a specific RMS across audio signals. Gain adjustments without explicit RMS targeting can result in inaccurate loudness levels especially those with large dynamic ranges.

## Potential Implementation
An RMS normalization feature would provide a more accurate way to achieve a set level across different audio signals. Considering the function already available in the library. It would be similar to the gain augmentation. The actual gain would be a function of the signal current level and of the target RMS value. 

## Considerations for Preventing Clipping
When setting high levels, clipping can occur if the signal's peak amplitude exceeds the system's maximum representable limit, leading to distortion. To prevent this issue, we could take into consideration the crest factor for different types of signals.

I believe this would enhance the library's utility for audio processing and production tasks. Thank you for considering this proposal. I look forward to discussing it further and am open to contributing to its implementation.

Shohail-Ismail:
### Summary
Adds a new `RMSNormalization` transform (subclass of `BaseWaveformTransform`) that:

- Converts a target RMS level in dBFS into a linear gain factor
- Computes the combined‐signal RMS, applies gain = target_amp / (rms + eps)
- Returns results in an `ObjectDict` (samples, sample_rate, etc.)

### Tests
Includes three basic tests covering:
1. Adjusting a 440 Hz sine wave from -20 dBFS to -10 dBFS
2. Leaving silent input unchanged
3. Applying equal gain to both channels of a stereo signal

### Issue
Closes issue #176

---
Let me know if you’d like any further tweaks, thanks!

### DIFF
--- a/tests/test_rms_normalization.py
+++ b/tests/test_rms_normalization.py
@@ -1,0 +1,58 @@
+import unittest
+import torch
+
+from torch_audiomentations import RMSNormalization
+
+# Allow some tolerance in dB comparisons due to numeric imprecision
+THRESHOLD_DB = 1.0
+
+class TestRMSNormalization(unittest.TestCase):
+    def test_sine_wave_adjustment(self):
+        """Transform should adjust a 440 Hz sine from -20 dBFS to -10 dBFS"""
+        sample_rate = 16000
+        t = torch.arange(0, sample_rate) / sample_rate
+        wave = torch.sin(2 * torch.pi * 440 * t)
+
+        # Scale to -20 dBFS
+        current_rms = torch.sqrt(torch.mean(wave.pow(2)))
+        target_amp = 10 ** (-20.0 / 20.0)
+        wave = wave * (target_amp / current_rms)
+
+        transform = RMSNormalization(-10.0)
+        result = transform(wave.unsqueeze(0).unsqueeze(0), sample_rate)
+
+        # Compute RMS of output
+        out_rms = torch.sqrt(torch.mean(result.samples.pow(2))).item()
+
+        # Manual RMS calculation for cross-check
+        flattened = result.samples.flatten()
+        total_power = sum(float(v) ** 2 for v in flattened)
+        rms_manual = (total_power / flattened.numel()) ** 0.5
+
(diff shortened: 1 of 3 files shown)


iver56 (changes requested):
Thanks for the PR!

For consistency with the other transforms, I would like RMSNormalization to also have/support `mode` and `p_mode` arguments.

Also, let `p` default to 0.5 for now, so it has the same default as the others. I realize that this default might not be ideal, but consistency is good. An overall change in the default `p` in the future will need to be a separate effort.

Shohail-Ismail:
Thanks for the feedback! Args mode and p_mode are now supported, and p defaults to 0.5 Let me know if there is anything further.

iver56:
Thanks for the update! I will have another look soon. I'm on vacation these days. That is what has delayed the review. Thanks for the patience

Shohail-Ismail:
Hi @iver56, I would appreciate it if you could take a look when you get the chance. Thanks!

iver56 (changes requested):
I ran `pytest`. It gave me an error:

```
torch_audiomentations\augmentations\rms_normalization.py:7: in <module>
    from ..utils import db_to_amplitude
E   ImportError: cannot import name 'db_to_amplitude' from 'torch_audiomentations.utils' (C:\Users\iver5\Code\torch-audiomentations\torch_audiomentations\utils\__init__.py)
```

Shohail-Ismail:
@iver56 Apologies, I'm not sure why this wasn't apparent in my pytest a month ago, but regardless I have made the changes and pytest finishes cleanly.

iver56:
Thanks! I will have a look

